PR #3, opened 2019-09-04
Rebuilding git logs all the live long day

kris-nova:
Fresh PR into master


kubernetes/audit.go
```diff
@@ -0,0 +1,114 @@
...
+		"--feature-gates=DynamicAuditing=true")
+	apiServer.Spec.Containers[0].Command = append(apiServer.Spec.Containers[0].Command,
+		"--runtime-config=auditregistration.k8s.io/v1alpha1=true")
+
+	//_, err = i.k8s.client.CoreV1().Pods(APIServerNamespace).Update(&apiServer)
+	//if err != nil {
+	//	return fmt.Errorf("unable to update API server: %v", err)
+	//}
+
+	// TODO - Generate YAML and patch the server manifest
```
kris-nova: There is an `ssh` package we use in `kubicorn` we could use to patch the API server manifest. 

We can at least generate the YAML here and wipe it, so we don't have to try to modify the existing one. We can just replace. 

AuditSink won't work until we fix this.

kubernetes/falco.go
```diff
@@ -0,0 +1,473 @@
...
+
+	// NamespaceName is an example of a "Global" variable we certainly should have
+	NamespaceName string
+
+	// DaemonSetName is an example of a "Possible" variable we don't know if we wan't to expose
+	DameonSetName string
+}
+
+// Delete is brute force method used to permanently destroy Falco resources in Kubernetes.
+func (i *FalcoInstaller) Delete(k8s *kubernetesConfigClient) error {
```
kris-nova: Right now we leave various resources installed (for instance the AuditSink) we probably want to delete every resource in backwards order

cmd/installFalco.go
```diff
@@ -0,0 +1,64 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	kubernetesfalc "github.com/kris-nova/falcoctl/kubernetes"
```
Issif:
```suggestion
	kubernetesfalc "github.com/falcosecurity/falcoctl/kubernetes"
```

main.go
```diff
@@ -0,0 +1,22 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package main
+
+import "github.com/kris-nova/falcoctl/cmd"
```
Issif:
```suggestion
import "github.com/falcosecurity/falcoctl/cmd"
```

cmd/delete.go
```diff
@@ -0,0 +1,46 @@
...
+	Short: "Delete a resource",
+	Long: ``,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Println("delete called")
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(deleteCmd)
+
```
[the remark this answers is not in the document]
kris-nova: I just went with what cobra did by default. I don't think this is a breaking change - but feel free to PR afterwords if you think this is a deal breaker.

ds.yaml
```diff
@@ -0,0 +1,62 @@
...
+            path: /proc
+        - name: boot-fs
+          hostPath:
+            path: /boot
+        - name: lib-modules
+          hostPath:
+            path: /lib/modules
+        - name: usr-fs
+          hostPath:
+            path: /usr
```
markyjackson-taulia: Can we add a line feed here

kris-nova: This file was for reference ONLY - also I don’t think a newline is syntacticly more valid - happy to rm this entire file if we want to block the PR on this

kris-nova: This file was for reference ONLY - also I don’t think a newline is syntacticly more valid - happy to rm this entire file if we want to block the PR on this

markyjackson-taulia: I am ok with resolving this conversation but generally speaking, since git was developed for linux  it is a good style to always put the newline as a last character if it is allowed by the file format. Furthermore, for example, for C and C++ header files it is required by the language standard. Although I totally get this is not either of those languages.
I am a fan of code style standardization and if the code base follows a pattern it should be adhered to. 
This also opens up the discussion for style guide or at least reference to such a guide.

PR #61, opened 2020-01-24
Rename probeloader to kernelmoduleloader

### LINKED ISSUE #53: Probe loader should be renamed
**What happened**:

Doing some homework on building an `eBPF` probe loader into `falcoctl` and I checked out the `probeloader` package.

It's currently using syscall `313` which is for loading kernel modules (not eBPF probes)

Maybe we should call the package `kernelmoduleloader`?

AnishShah:
**What type of PR is this?**

/kind cleanup

/area library

Cleanup

Fixes #53 

```release-note
chore(cmd): rename falcoctl install probe to falcoctl install module
```


cmd/install_probe_linux.go
```diff
@@ -108,15 +108,15 @@ func NewProbeInstallCommand(streams genericclioptions.IOStreams) *cobra.Command
 			}
 
 			// fetch module
-			err = probeloader.FetchModule(o.falcoProbeURL, falcoProbeFullpath)
+			err = kernelmoduleloader.FetchModule(o.falcoProbeURL, falcoProbeFullpath)
```
kris-nova: Looks like we are still using the word `probe` in this package. 

We need to be very clear about when we are referring to a kernel module, or a BPF probe - they are very different things

AnishShah: Sorry Kris. I didn't understand. You mean, refactoring loader.go file such that we separate functions related to BPF probe and kernel module into respective pkgs?

kris-nova: So there are two things, that require completely different logic.

1) Loading a kernel module such that Falco can use it to consume kernel events
2) Loading a BPF probe such that Falco can use it to consume kernel events

They are very different things, and (e)BPF is significantly more stable and reliable

---

Renaming the package is a good step in the broader effort here. 

We should only use the word `probe` if we are referring to anything related to (e)BPF

We should only use the word `module` if we are referring to anything related to a kernel module 

The `install_probe_linux.go` file is a misnomer. It does **not** install a BPF probe, it installs a kernel module (kind of). 

Having a file named `install_probe_linux.go` that installs a kernel module makes about as much sense as having a file named `install_emacs_linux.go` that installs actually installs `vim`. 

---

> Sorry Kris. I didn't understand. You mean, refactoring loader.go file such that we separate functions related to BPF probe and kernel module into respective pkgs?

Yes. Please have two packages. One for loading a BPF probe, and one for loading a kernel module. They are different things and require different logic, and should be named differently.

kris-nova: A rule of thumb here is that if we are dealing with a kernel module and the word `probe` appears -- there is a significant chance something is named incorrectly.

kris-nova: In order to close #53 the word `probe` should be completely removed from the codebase. 

Right now `falcoctl` doesn't do any `probe` loading. It however does `kernel module` loading.

If you want to refactor the word out, please do and we can merge this in :smile: 

Sorry to be such a pain - but there is a **LOT** of misnomers in Falco, and we have to start being very diligent about how we name things.

AnishShah: Thanks for the explanation, Kris. So we should also rename `install probe` command to `install module`.

cmd/install_module_linux.go
```diff
@@ -0,0 +1,136 @@
...
+				return err
+			}
+
+			logger.Always("FALCO_VERSION: %s", o.falcoVersion)
+			logger.Always("FALCO_MODULE_URL: %s", o.falcoModuleURL)
+			logger.Always("FALCO_MODULE_REPO: %s", o.falcoModuleRepo)
+			logger.Always("KERNEL_VERSION: %s", falcoKernelRelease)
+			logger.Always("KERNEL_CONFIG_HASH: %s", falcoConfigHash)
+
+			// if FALCO_PROBE_URL not set, build it
```
kris-nova: Can we adjust or delete this comment please

kris-nova:
```suggestion
			// if FALCO_MODULE_URL not set, build it
```

pkg/kernelmoduleloader/loader.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
kris-nova: Can we also rename this file from `loader.go` to `kernel_module_loader.go`?

We will more than likely have one for `bpf_probe_loader.go` in the future

kris-nova:
Looks like we need a rebase :disappointed:

AnishShah:
Hi Kris. I have rebased my branch.

AnishShah:
Hi Kris. Any idea how to merge this? It says that it is in a merge pool

PR #70, opened 2020-01-30
Make AuditSink creation idempotent

### LINKED ISSUE #47: Creation of AuditSink isn't idempotent
**What happened**:

According to the code comments [here](https://github.com/falcosecurity/falcoctl/blob/master/pkg/kubernetes/falco.go#L96-97):
```
<snip>
// Install is an idempotent installer for Falco in Kubernetes. By design if a resource already exists, we log it
// and move on with installing the others.
</snip>
```
However I get the following error if the AuditSink already exists:

`Fatal error: auditsinks.auditregistration.k8s.io "falco-audit-sink" already exists`

I would expect to receive a message similar to the others when they already exist.. something like:

`AuditSink already exists: falco-audit-sink`

```
falcoctl install falco
falcoctl install falco
```

After the second run, you should see the error message

ndhanushkodi:
Create AuditSink idempotently to fix #47 

**What type of PR is this?**

/kind bug

/area library

/area cli

It creates the AuditSink if it doesn't exist, and updates when it does. It also refactors to use the apierrors package for error checking.

Fixes #47

Added small refactor to use apierrors package to reduce hardcoding for error checking

Added a log line that to the output that is user-facing

```release-note
fix: Creation of AuditSink isn't idempotent (#47)
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/mattn/go-isatty v0.0.10 // indirect
 	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
-	github.com/pkg/errors v0.8.1 // indirect
+	github.com/pkg/errors v0.8.1
 	github.com/spacemonkeygo/openssl v0.0.0-20181017203307-c2dcc5cca94a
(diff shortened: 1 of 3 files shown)


mstemm:
I'd like to defer to @leodido, as I haven't worked on this part of falcoctl.

Issue #71, opened 2020-02-05
Add command to obtain Falco version


kris-nova:
**What would you like to be added**:

```
falcoctl version
```

This would similar to `kubectl version` and work with this new API https://github.com/falcosecurity/falco/pull/872

**Why is this needed**:

abd2312:
Hi @kris-nova  , I want to work on this if it's okay with you.

leodido:
Hey Kris, this should be something like (imho): 

```console
falcoctl falco version
```

This because the Falco Version API returns the Falco version.

RajatVaryani:
@leodido @kris-nova Is this open for contribution?

PR #74, opened 2020-02-07
falcoctl version command

saiyam1814:
Signed-off-by: Saiyam Pathak <saiyam911@gmail.com> , Sarwesh Suman 
                            

This PR is for falcoctl version addressing issue #71 
/kind feature 

```release-note
falcoctl version
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -90,6 +90,7 @@ func New(streams genericclioptions.IOStreams) *cobra.Command {
 	cmd.AddCommand(Install(streams, f))
 	cmd.AddCommand(Delete(streams, f))
 	cmd.AddCommand(Convert(streams))
+        cmd.AddCommand(MakeVersion(streams))
 
 	return cmd
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,46 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"os/exec"
+	"strings"
+	"github.com/spf13/cobra"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
+)
+
+//Default version if tag not available
+var (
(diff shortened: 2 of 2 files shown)


cmd/root.go
```diff
@@ -90,6 +90,7 @@ func New(streams genericclioptions.IOStreams) *cobra.Command {
 	cmd.AddCommand(Install(streams, f))
 	cmd.AddCommand(Delete(streams, f))
 	cmd.AddCommand(Convert(streams))
+        cmd.AddCommand(MakeVersion(streams))
```
Issif: little issue with tabulation

cmd/version.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+//Default version if tag not available
+var (
+	Version string = "Not Available"
+)
+
+//Getting the latest GitHub Version from tag
+func getVersion() string {
+	cmd := exec.Command("git", "describe", "--abbrev=0", "--tags")
```
Issif: It implies to have `git` locally, can we think about making an http call to github API? https://developer.github.com/v3/git/tags/

saiyam1814:
> Hey @saiyam1814, the title of that issue is misleading...
> The title says "Add command to obtain Falco version", but the issue corpus seems to refer to `falcoctl` version. They are two different things, clearly.
> 
> It's not your fault, thanks for submitting this PR! :)
> 
> So, in case, do you want, with this PR, implement `falcoctl version` I have some comments:
> 
> * do not depend on `git`
> * use `LDFLAGS` during the compilation of `falcoctl` to set a version constant
> * return that constant from the `falcoctl version` command
> * remove reference from this PR to issue #71, that, as said, refers to `falcoctl falco version` :)

Hi @leodido  Sure thanks  I will do the changes

PR #85, opened 2020-03-26
flags fallback logic and general improvements

leogr:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

This is a further PR including a bunch of improvements:
- [x] Refactored config option with config file support
- [x] Kube flags are attached to commands only when needed
- [x] Initial setup for options validation
- [x] Clean up unnecessary dependencies in options and constructors
- [x] Automatically fallback all flags to ENV and config file
- [x] Bump dependencies

Fixes #

This PR also introduces an important (but maybe opinionated) feature which is described below.

When a flag is not provided by the user, fallback to one of (in order of precedence):
- ENV (with `FALCOCTL_` prefix)
- config file (e.g. `~/.falcoctl.yaml`)
- its default

This PR introduces the logic above for all flags. If we did not agree on this, alternatively we could use it for persistent flags only or selectively (but I would avoid this, because that will make the code more complex and increase the maintenance cost, IMHO). 

```release-note
new(cmd): automatically fallback all flags to ENV and config file
```


cmd/convert.go
```diff
@@ -33,9 +32,7 @@ func (o ConvertOptions) Validate(c *cobra.Command, args []string) error {
 
 // NewConvertOptions instantiates the `convert` command options
 func NewConvertOptions(streams genericclioptions.IOStreams) CommandOptions {
```
leodido:
```suggestion
func NewConvertOptions() CommandOptions {
```

PR #92, opened 2020-04-06
CLI tests

### LINKED ISSUE #87: CLI tests

Since `falcoctl` starts to grow, it would be a good thing to also test its CLI.

This can be done in the same way I did for `driverkit`, [here](https://github.com/falcosecurity/driverkit/pull/49).

**Why is this needed**:

The goal is to be able to introduce new flags, new sub-commands, etc. without breaking the existing ones.

leodido:
**What type of PR is this?**

/kind feature

/area cli

/area tests

Setups tests for CLI.

Fixes #87 

```release-note
NONE
```

### DIFF
--- a/cmd/cli_test.go
+++ b/cmd/cli_test.go
@@ -1,0 +1,107 @@
+// +build !race
+
+package cmd
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/acarl005/stripansi"
+	"gotest.tools/assert"
+)
+
+type expect struct {
+	err string
+	out string
+}
+
+type testCase struct {
+	descr  string
+	env    map[string]string
+	args   []string
+	expect expect
+}
+
+var tests = []testCase{
+	{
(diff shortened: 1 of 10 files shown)


cmd/cli_test.go
```diff
@@ -0,0 +1,120 @@
+// +build !race
```
leogr, line 1: @leodido as discussed privately, we can remove this and temporarily disable the `--race` option in the make file, in order to make this test run on the CI. 
I will do it.

cmd/root.go
```diff
@@ -32,13 +33,29 @@ import (
...
+	}
+}
+
 func init() {
 	configOptions = NewConfigOptions()
 	cobra.OnInitialize(initConfig)
 }
 
+// FalcoctlRoot wraps the root cobra.Command.
+type FalcoctlRoot struct {
```
leogr, line 53: I believe we can avoid this wrapper, because the same result can be achieved using `cobra.Command` directly.
[Here](https://github.com/leogr/event-generator/blob/master/cmd/root.go) an example.

leogr:
LGTM

Issue #110, opened 2022-06-07
Apple M1 / darwin-arm64 build


shadiramadan:
**What would you like to be added**:
Apple M1 support via a darwin-arm64 build.

**Why is this needed**:
Apple M1 is arm based.

leogr:
cc @alacuku

alacuku:
Hi @shadiramadan,
the `darwin-arm64` build is available here: https://github.com/falcosecurity/falcoctl/actions/runs/2835316120. Note that the artifacts are retained just for 24 hours. 
Anyway, you should be able to build `falcoctl` from the source. Another alternative could be to wait for the next release.

cpanato:
this is done already
/close

PR #122, opened 2022-08-10
update(cli): update install tls command

alacuku:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

This PR refactors the `install tls` command. The idea is to have only one version of the command for all the supported operating systems such as `linux, macOS and windows`.

What this PR does is:
* keep only one version of the command by removing the one for linux;
* Move the logic of the command + tls generator under pkg/install/tls;
* Do not use the `CommandOptions` interface which makes cumbersome handling options for each command instead the options are handled by the `NewInstallTLSCmd()` function. Furthermore, the `CommandOptions` interface does not add any benefit. 
* The cert files are saved under the working directory of falcoctl by default. The user can ovverride by passing the path to a target directory.

Falcoctl now builds for also for macOS and windows in addition to linux. #110 

Fixes #


leogr (approved):
LGTM

PR #131, opened 2022-09-05
Implementation of a skeleton of OCI registry related commands

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

This PR aims at implementing the first skeleton of the `login` `logout` `push` and `pull` commands. It does so by using ORAS as an underlying library, making `falcoctl` able to interact with registries compliant with OCI specifications (artifact-spec and distribution-spec). This is the first step towards a renovated `falcoctl`, so feedback is highly appreciated! Here we can also discuss what options to add in the future that could be useful to the user or our CI.
The implemented commands were successfully tested against the Github Container Registry (GHCR). In the following PRs, we will also add some tests on the newly introduced library to prove its correctness. I will also take care to open an issue for tracking down the evolution of the new `falcoctl`, considering all the relevant tasks.

Fixes #


leogr (commented):
SGTM.

I'm just wondering if there's a specific motivation for the `ConfigOptions` to `CommonOptions` renaming :thinking: 

I would have left it as-is, but perhaps am I missing something?

loresuso:
I thought about changing it because CommonOptions in the future will contain options related to all the commands. In this way, we will have composite options using also newly defined structs such as the options needed for interacting remotely with the registry (for instance plain-http, ca-cert and so on). WDYT? :)

LucaGuerra:
This is great :rocket: 

I have a question about the `pull` command: what does pull mean in this context? Will it get the artifact and copy it somewhere? Does it require a directory where to copy the necessary file? Can or should the user specify the directory?

I might be biased since I think about pull in the context of "docker pull" which just copies the thing in its own internal storage but the context is different here :sweat_smile:

loresuso:
@LucaGuerra The `pull` command for now just limits itself to downloading the artifacts in the current directory. In the next PRs we will add the option to let the user specify where to download the artifacts. We preferred doing it this way since this PR aims at building the first skeleton of the new `falcoctl` 🙂

LucaGuerra:
@loresuso makes sense! The next step is great for next PRs :muscle:

leogr (approved):
LGTM :+1:

PR #159, opened 2022-09-22
new: implement artifact install command

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

This PR implements the `artifact install` command.  This command is basically implemented starting from the already available operations on index files and the functionalities offered by the puller. 

Fixes #

### DIFF
--- a/cmd/artifact_install.go
+++ b/cmd/artifact_install.go
@@ -81,6 +81,10 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 		return err
 	}
 
+	if len(mergedIndexes.Entries) < 1 {
+		o.Printer.Warning.Println("No configured index. Consider to configure one using the 'index add' command.")
+	}
+
 	o.credentialStore, err = authn.NewStore([]string{}...)
 	if err != nil {
 		return err
@@ -137,7 +141,7 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 		// Extract artifact and move it to its destination directory
 		err = utils.ExtractTarGz(f, destDir)
 		if err != nil {
-			return err
+			return fmt.Errorf("cannot extract %q to %q: %w", result.Filename, destDir, err)
 		}
 
 		err = os.Remove(result.Filename)
--- a/cmd/artifact_install.go
+++ b/cmd/artifact_install.go
@@ -1,0 +1,192 @@
+// Copyright 2022 The Falco Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"github.com/falcosecurity/falcoctl/cmd/internal/utils"
+	"github.com/falcosecurity/falcoctl/pkg/index"
+	"github.com/falcosecurity/falcoctl/pkg/oci"
+	"github.com/falcosecurity/falcoctl/pkg/oci/authn"
+	ocipuller "github.com/falcosecurity/falcoctl/pkg/oci/puller"
+	"github.com/falcosecurity/falcoctl/pkg/options"
+)
+
+const (
+	defaultPluginsDir    = "/usr/share/falco/plugins"
+	defaultRulesfilesDir = "/etc/falco"
+)
+
+type artifactInstallOptions struct {
+	*options.CommonOptions
+	credentialStore *authn.Store
+	rulesfilesDir   string
+	pluginsDir      string
+}
+
+// NewArtifactInstallCmd returns the artifact search command.
+func NewArtifactInstallCmd(ctx context.Context, opt *options.CommonOptions) *cobra.Command {
+	o := artifactInstallOptions{
+		CommonOptions: opt,
+	}
+
+	cmd := &cobra.Command{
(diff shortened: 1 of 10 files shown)


cmd/artifact_install.go
```diff
@@ -0,0 +1,192 @@
...
+		if err != nil {
+			return fmt.Errorf("cannot load index %s: %w", i.Name, err)
+		}
+		allIndexes = append(allIndexes, i)
+	}
+
+	o.Printer.Info.Println("Merging all configured indexes ...")
+	mergedIndexes := index.NewMergedIndexes()
+	mergedIndexes.Merge(allIndexes...)
+	o.Printer.Verbosef("All configured indexes have been merged: %d", len(allIndexes))
```
leogr, line 97: I guess this will be useful for the search command too.

cmd/artifact_install.go
```diff
@@ -0,0 +1,192 @@
...
+		if strings.ContainsAny(name, ":@") {
+			ref = name
+		} else {
+			entry, ok := mergedIndexes.EntryByName(name)
+			if !ok {
+				o.Printer.Warning.Printf("cannot find %s among the configured indexes, skipping\n", name)
+				continue
+			}
+			ref = fmt.Sprintf("%s/%s:latest", entry.Registry, entry.Repository)
+		}
```
leogr, line 123: I believe we need a simple parser for this.

cmd/artifact_install.go
```diff
@@ -0,0 +1,192 @@
...
+			return err
+		}
+
+		var destDir string
+		switch result.Type {
+		case oci.Plugin:
+			destDir = o.pluginsDir
+		case oci.Rulesfile:
+			destDir = o.rulesfilesDir
+		}
```
leogr, line 149: Destination dir should be configurable via command line flags.

pkg/oci/puller/puller.go
```diff
@@ -56,14 +58,23 @@ func (p *Puller) Pull(ctx context.Context, artifactType oci.ArtifactType, ref, d
...
 	copyOpts := oras.CopyOptions{}
 	copyOpts.Concurrency = 1
-	if artifactType == oci.Plugin {
+	if rootDesc.MediaType == v1.MediaTypeImageIndex {
 		plt := &v1.Platform{
 			OS:           os,
 			Architecture: arch,
 		}
 		copyOpts.WithTargetPlatform(plt)
+		artifactType = oci.Plugin
```
leogr, line 75: We should get this from the descriptor, not just assuming that if a fat manifest then it's a plugin.

leogr (commented):
Look great! I have just left some comments.

pkg/oci/puller/puller.go
loresuso: Completely agreed!

cmd/artifact_install.go
loresuso: I have added a dedicated logic for this

loresuso:
This PR is now ready for another round of review!

loresuso:
Another question I have is that the `artifact install` command relies on the fact that the user was already logged in to each registry artifact may come from. I think that this is legit since the install command most likely is preceded by a `artifact search`, where the user can retrieve info about an artifact and most importantly the registry hosting it. This is what I think but I would like to hear other opinions on this

cmd/internal/utils/extract.go
```diff
@@ -0,0 +1,83 @@
...
+			if err != nil {
+				return err
+			}
+			if err := copyInChunks(outFile, tarReader); err != nil {
+				return err
+			}
+			err = outFile.Close()
+			if err != nil {
+				return err
+			}
```
alacuku, line 61:
```suggestion
			if err = outFile.Close(); err != nil {
				return err
			}
```

alacuku (commented):
Hi @loresuso, nice PR! 
I left some comments/suggestions.

alacuku:
> Another question I have is that the `artifact install` command relies on the fact that the user was already logged into each registry artifact may come from. I think that this is legit since the install command most likely is preceded by a `artifact search`, where the user can retrieve info about an artifact and most importantly the registry hosting it. This is what I think but I would like to hear other opinions on this

I've just tested it and it works in the case when the user is not authenticated into the remote repository as far as the remote repo is public!

cmd/artifact_install.go
```diff
@@ -0,0 +1,170 @@
...
+		DisableFlagsInUseLine: true,
+		Short:                 "Install a list of artifacts",
+		Long:                  "Install a list of artifacts",
+		Args:                  cobra.MinimumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			o.Printer.CheckErr(o.RunArtifactInstall(ctx, args))
+		},
+	}
+
+	cmd.Flags().StringVarP(&o.rulesfilesDir, "rulesfiles-dir", "", defaultRulesfilesDir,
```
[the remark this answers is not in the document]
loresuso: If we use only one flag, what happens if the same install command is used to install both a plugin and a rule at the same time? I think it is better to keep these flags separated also because when we will be able to handle dependencies (subsequent PRs), we will have ways to say where artifacts should be moved independently. But I am open to modify this if there are other reasons to do that!

cmd/artifact_install.go
```diff
@@ -0,0 +1,170 @@
...
+
+		f, err := os.Open(result.Filename)
+		if err != nil {
+			return err
+		}
+
+		// Extract artifact and move it to its destination directory
+		err = utils.ExtractTarGz(f, destDir)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
loresuso: Yep I agree. I am going to add this, but at the end of this first round of the functional requirements of falcoctl we will have a tour together to check every error message properly 🙂

alacuku (commented):
/lgtm

cmd/artifact_install.go
```diff
@@ -0,0 +1,174 @@
...
+)
+
+type artifactInstallOptions struct {
+	*options.CommonOptions
+	credentialStore *authn.Store
+	rulesfilesDir   string
+	pluginsDir      string
+}
+
+// NewArtifactInstallCmd returns the artifact search command.
```
leogr, line 46: Search?

leogr (approved):

PR #163, opened 2022-10-03
update: remove type from pull operations

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

This PR removes the need to specify a type when pulling artifacts. It is not necessary since we can infer the type by using media types. 
This PR will unblock the one opened in wip for the `install` command.

Fixes #

### DIFF
--- a/pkg/oci/puller/puller.go
+++ b/pkg/oci/puller/puller.go
@@ -95,6 +95,7 @@ func (p *Puller) Pull(ctx context.Context, ref, destDir, os, arch string) (*oci.
 	if err != nil {
 		return nil, fmt.Errorf("unable to fetch descriptor with digest %q: %w", desc.Digest, err)
 	}
+	defer descReader.Close()
 
 	var manifest v1.Manifest
 	descBytes, err := io.ReadAll(descReader)
--- a/pkg/oci/puller/puller.go
+++ b/pkg/oci/puller/puller.go
@@ -91,7 +91,7 @@ func (p *Puller) Pull(ctx context.Context, ref, destDir, os, arch string) (*oci.
 			repo.Reference.Repository, repo.Reference.Reference, repo.Reference.Repository, err)
 	}
 
-	descReader, err := repo.Fetch(ctx, desc)
+	descReader, err := localTarget.Fetch(ctx, desc)
 	if err != nil {
 		return nil, fmt.Errorf("unable to fetch descriptor with digest %q: %w", desc.Digest, err)
 	}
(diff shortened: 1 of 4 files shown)


pkg/oci/puller/puller.go
```diff
@@ -83,7 +91,37 @@ func (p *Puller) Pull(ctx context.Context, artifactType oci.ArtifactType, ref, d
 			repo.Reference.Repository, repo.Reference.Reference, repo.Reference.Repository, err)
 	}
 
+	descReader, err := repo.Fetch(ctx, desc)
+	if err != nil {
+		return nil, fmt.Errorf("unable to fetch descriptor with digest %q: %w", desc.Digest, err)
+	}
+
```
alacuku, line 98:
```suggestion
defer descReader.Close()
```

pkg/oci/puller/puller.go
```diff
@@ -83,7 +91,37 @@ func (p *Puller) Pull(ctx context.Context, artifactType oci.ArtifactType, ref, d
 			repo.Reference.Repository, repo.Reference.Reference, repo.Reference.Repository, err)
 	}
 
+	descReader, err := repo.Fetch(ctx, desc)
```
alacuku, line 94:
```suggestion
	descReader, err := localTarget.Fetch(ctx, desc)
```
We have already fetched the graph from the remote repo, we can get the descriptor from the local store to avoid fetching it from the remote repo.

pkg/oci/types.go
```diff
@@ -61,6 +61,7 @@ func (e *ArtifactType) Type() string {
 type RegistryResult struct {
 	Digest string
 	Config ArtifactConfig
+	Type   ArtifactType
```
alacuku, line 64: Should we add the platform, for the artifacts that support it, to the `RegistryResult`? If yes it should be easy to add such information when pulling artifacts.

alacuku (commented):
Hi @loresuso, thanks for the PR. 
Left some comments and suggestions.

pkg/oci/puller/puller.go
loresuso: Yes, you are super right!

pkg/oci/types.go
loresuso: `os` and `arch` are params of the `Pull` function and also for the `Push` we have a way to specify them with options, so the caller already knows which platform is going to be used for artifacts that support it. If a wrong platform is passed for example in the pull operation, we are printing out this error: `no matching manifest was found in the manifest list` deriving from the `SelectManifest` function in oras

alacuku (commented):
/LGTM

leogr (approved):

PR #167, opened 2022-10-05
update: improve artifact search command

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

Improve artifact search command by using Levenshtein Distance and partial matching on keywords.
This allows returning meaningful results also if a name is e.g. mistyped or we have only a partial match on keywords (for instance `aws_cloudtrail` keyword can now match with `aws`). To implement the Levenshtein distance a DP matrix is used. 
Future PR will also add a way to look up using regular expressions. 

Fixes #

### DIFF
--- a/cmd/artifact_search.go
+++ b/cmd/artifact_search.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	minScore = 0.65
+	defaultMinScore = 0.65
 )
 
 type artifactSearchOptions struct {
@@ -63,7 +63,7 @@ func NewArtifactSearchCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 		},
 	}
 
-	cmd.Flags().Float64VarP(&o.minScore, "min-score", "", minScore,
+	cmd.Flags().Float64VarP(&o.minScore, "min-score", "", defaultMinScore,
 		"the minimum score used to match artifact names with search keywords")
 
 	return cmd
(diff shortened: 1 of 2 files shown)


FedeDP:
Very cool!

cmd/artifact_search.go
```diff
@@ -25,8 +26,21 @@ import (
 	"github.com/falcosecurity/falcoctl/pkg/output"
 )
 
+const (
+	minScore = 0.65
```
FedeDP, line 30: Perhaps
```suggestion
	defaultMinScore = 0.65
```
?

pkg/index/index.go
```diff
@@ -188,27 +190,95 @@ func (m *MergedIndexes) Merge(indexes ...*Index) {
...
-	var result []*Entry
-	keywordSet := make(map[string]bool)
-
-	for _, keyword := range keywords {
-		keywordSet[keyword] = true
-	}
+// minScore is the minimum score to consider a match between a name of an artifact and a keyword.
+// if minScore is not reached, we fallback to a simple partial matching on keywords.
+func (i *Index) SearchByKeywords(minScore float64, keywords ...string) []*Entry {
+	var matches []*Entry
```
FedeDP, line 196: Can't we use a  set of string here (ie: map[string]struct{}) and then return its keys, instead of deduplicating keys below?

loresuso: What do you think about it now? :)

FedeDP (approved):
/approve

PR #169, opened 2022-10-07
chore: rename `install tls` command to `tls install` to be compliant with new naming convention

loresuso:
This allow to be compliant with the naming we have chosen for this repo and the other commands.

Signed-off-by: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

The naming convention for command is object and then verb, i.e `artifact install`.
To respect this, change `install tls` with `tls install`.

Fixes #

### DIFF
--- a/cmd/tls.go
+++ b/cmd/tls.go
@@ -18,17 +18,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// NewInstall creates the `install` command
-func NewInstallCmd() *cobra.Command {
+// NewTLSCmd return the tls command.
+func NewTLSCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:                   "install",
+		Use:                   "tls",
 		TraverseChildren:      true,
 		DisableFlagsInUseLine: true,
-		Short:                 "Install a component with falcoctl",
-		Long:                  `Install a component with falcoctl`,
+		Short:                 "Generate and install TLS material for Falco",
+		Long:                  `Generate and install TLS material for Falco`,
 	}
 
-	cmd.AddCommand(NewInstallTLSCmd())
+	cmd.AddCommand(NewTLSInstallCmd())
 
 	return cmd
 }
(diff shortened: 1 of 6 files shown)


cmd/tls.go
```diff
@@ -18,17 +18,17 @@ import (
...
+// NewTLSCmd return the tls command.
+func NewTLSCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:                   "install",
+		Use:                   "tls",
 		TraverseChildren:      true,
 		DisableFlagsInUseLine: true,
-		Short:                 "Install a component with falcoctl",
-		Long:                  `Install a component with falcoctl`,
+		Short:                 "Handle certificates with falcoctl",
```
loresuso, line 27: Is this a right choice for the description? Can we find better?

FedeDP:
> To respect this, change install tls with tls install.

Just out of curiosity: isn't `tls` more like flag of the  `artifact install` command?

loresuso:
> > To respect this, change install tls with tls install.
> 
> Just out of curiosity: isn't `tls` more like flag of the `artifact install` command?

hey @FedeDP, more than a flag `tls` could be a subcommand. I think this is a matter of taste, but since we have chosen the pattern `NOUN` `VERB` for the other commands, I now found having `install tls` a bit misleading, especially because of the `artifact install` command

FedeDP:
> but since we have chosen the pattern NOUN VERB for the other commands

Makes sense; but are we really installing a `tls` ? :)

cmd/tls.go
leogr: My proposal is: `Generate and install TLS material for Falco`

leogr:
Hey @FedeDP @loresuso 

Yes, it actually installs the TLS material (in a sense indicated by the description: `This command is a convenience to not only generate the TLS material, but also drop it off on the local filesystem.`).

@loresuso this PR will break that blog post anyway, so I agree that the less we change, the better, but we still have to indicate this as a breaking change and maybe update that blog post too.


PS
left only one suggestion, otherwise SGTM

leogr (approved):
LGTM

PR #187, opened 2022-12-01
update(internal/utils): return full path of the extracted files in ExtractTarGz func

alacuku:
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

/area library

> /area cli

> /area tests

> /area examples

Return full path of the extracted files in ExtractTarGz func. Having the path of the extracted files simplifies subsequent operations like moving around the files if we extract them in a temporary dir.

Fixes #

### DIFF
--- a/internal/artifact/install/install.go
+++ b/internal/artifact/install/install.go
@@ -127,7 +127,7 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 		}
 
 		// Extract artifact and move it to its destination directory
-		err = utils.ExtractTarGz(f, destDir)
+		_, err = utils.ExtractTarGz(f, destDir)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/artifact/install/install.go
```diff
@@ -127,7 +127,7 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 		}
 
 		// Extract artifact and move it to its destination directory
-		err = utils.ExtractTarGz(f, destDir)
+		_, err = utils.ExtractTarGz(f, destDir)
```
cpanato, line 130: are we planning to use the return result?

alacuku: Yes, in the upcoming PR we use it!

PR #195, opened 2022-12-17
refactor: create registry package and port oauth and plain http flags to artifact subcommands

loresuso:
**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

This PR tries to better organize the code by introducing the registry package. This is pretty specular to the repository one, and it is useful for implementing and centralizing operations that might be performed against a remote registry, like checking if the latter implements the API that we need to push and pull artifacts. It is better in this way also because we can have the client injected from the outside, which is good also for testability and clearness, given that we do the same for the repository. 

Furthermore, remove some left over from previous refactors, like 
- creating the client in the right way in `artifact info`.
- move the responsibility of listing tags to the repository struct.

After doing that, it was easy to port OAuth and Plain HTTP flags to all the `artifact` subcommands that need them. 

Fixes #


internal/utils/validate.go
```diff
@@ -113,61 +105,3 @@ func ParseReference(mergedIndexes *index.MergedIndexes, name string) (string, er
...
-		return err
-	}
-
-	sp.Success(fmt.Sprintf("Remote registry %q implements docker registry API V2", regName))
-	printer.Verbosef("Proceeding as user %q", cred.Username)
-
-	return nil
-}
-
-func checkRegistryUnauthenticated(ctx context.Context, regName string) error {
```
loresuso, line 150: From my experiments, this function was useless, since if we have EmptyCredentials, no bearer is appended as authorization header, which is totally equivalent to what CheckRegistryConnection will do with empty credentials.

alacuku: I think we still need it. If the user is not authenticated commands interacting with a remote registry will fail with error: ` unexpected status code 403: denied: requested access to the resource is denied`

loresuso: I am thinking about this again. Since this function is always called in the flow for push and pull, for me it doesn't make sense checking the connection. The HTTP error we will receive from the first request for push and pull will exactly be the same. It makes much more sense, instead, when checking credentials for login: if we can connect to registry through the supplied credentials, then we save them. This is also the same thing that oras-cli does, the ping is only performed in the login, not at every push and pull operation. That's why I am deleting the check for ClientForRegistry function.

internal/registry/login/login.go
```diff
@@ -84,9 +85,15 @@ func (o *loginOptions) RunLogin(ctx context.Context, args []string) error {
...
+	client := authn.NewClient(authn.WithCredentials(cred))
+	r, err := registry.NewRegistry(reg, registry.WithClient(client))
+	if err != nil {
+		return fmt.Errorf("unable to create registry: %w", err)
+	}
+
+	if err := r.CheckConnection(ctx); err != nil {
 		o.Printer.Verbosef("%s", err.Error())
-		return fmt.Errorf("unable to connect to registry %q: %w", registry, err)
+		return fmt.Errorf("unable to connect to reg %q: %w", reg, err)
```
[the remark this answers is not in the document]
loresuso: Yes, thank you!

internal/artifact/info/info.go
```diff
@@ -92,16 +90,21 @@ func (o *artifactInfoOptions) RunArtifactInfo(ctx context.Context, args []string
...
 		if err != nil {
 			return err
 		}
 
-		client := authn.NewClient(authn.WithCredentials(&cred))
+		repo, err := repository.NewRepository(ref,
+			repository.WithClient(client),
+			repository.WithPlainHTTP(o.PlainHTTP))
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
loresuso: NewRepository method should already contain meaningful information

internal/utils/registry.go
```diff
@@ -65,20 +66,26 @@ func ClientForRegistry(ctx context.Context, registry string, oauth bool, printer
...
 		authn.WithOauth(oauth),
-		authn.WithCredentials(&cred)), err
+		authn.WithCredentials(&cred))
+
+	r, err := registry.NewRegistry(reg, registry.WithClient(client), registry.WithPlainHTTP(plainHTTP))
+	if err != nil {
+		return nil, fmt.Errorf("unable to create registry: %w", err)
+	}
+
+	if err := r.CheckConnection(ctx); err != nil {
```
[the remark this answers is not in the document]
loresuso: Yes, move discussion on that comment

internal/utils/validate.go
loresuso: The check is back, ported into the registry package and also improved by verifying correct header is set. This was needed to perform the additional check on the implementation of v2 (the above comment addressed only authentication). Thank you!

leogr (approved):
Nice :+1:

PR #222, opened 2023-01-19
new(internal/artifact): added support to pass cron-like strings to specify artifact follow interval

FedeDP:
**What type of PR is this?**

/kind feature

/area cli

Adds support for `cron` artifact follow option to pass a cron string; to be used in place of `every`.
`every` option remains supported; one cannot specify both flags together.

Fixes #


FedeDP:
Rebased on top of new main.

FedeDP:
I left the default value only for `every`, meaning that `cron` is disabled by default; we could switch them later if we wish.

PR #224, opened 2023-01-22
chore(falcoctl): bump oras-go to v2(GA)

arukiidou:
Signed-off-by: junya koyama <arukiidou@yahoo.co.jp>

**What type of PR is this?**

/kind cleanup

/area library

/area cli

- updase [oras-go/v2](https://github.com/oras-project/oras-go),
  - use GA

Fixes #

- Done.
  - run `go mod tidy`
  - passed `golangci-lint run --new-from-rev main`
  - passed `go test ./...`

- Always pushed as [OCI Image Manifest](https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest). 
- Currently, Manifests are not pushed as [OCI Artifact Manifest](https://github.com/opencontainers/image-spec/blob/main/artifact.md#oci-artifact-manifest-specification) yet.


arukiidou:
I'll fix it soon.

loresuso:
Hello @arukiidou and thanks for this effort! 
We are close to a release, I will be back right after to try to help you solve this problem, for sure switching to v2 is needed :)

arukiidou:
> Hello @arukiidou and thanks for this effort! We are close to a release, I will be back right after to try to help you solve this problem, for sure switching to v2 is needed :)

Sorry for the delay.
I am having a hard time with the destructive changes in v2.0.0-rc.4.
If you want to bring in regular contributors to 0.4 in time, feel free to close it.

arukiidou:
I think ready for the review.

pkg/oci/pusher/pusher.go
```diff
@@ -291,15 +295,20 @@ func (p *Pusher) packManifest(ctx context.Context, fileStore *file.Store,
 	// Now we can create manifest, using the Config descriptor and principal Layer descriptor.
 	// In case annotation source is passed, we put it in the ManifestAnnotations.
 	var packOptions oras.PackOptions
+
+	// Currently, Manifests are not pushed as OCI Artifact Manifest.
+	// Always pushed as OCI Image Manifest.
+	packImageManifest := true
```
loresuso, line 301: Nice that we can have this option so that we can plan when to switch to OCI Artifact Manifests later (maybe if we will consider using the `subject` fields, potentially for signing).

loresuso (commented):
Overall looks good! Need to find some time to properly test it before leaving my approval, thank you!

pkg/oci/pusher/pusher.go
maxgio92: This way we should expect a consistent behaviour with `packImageManifest=true` and switch when we want 😍

arukiidou: I agree.
But it is a bit difficult to me, designing and testing compatibility mode for ORAS,
Only bump will done here, and add this comment instead.

https://github.com/oras-project/oras/blob/v1.0.0-rc.1/docs/proposals/compatibility-mode.md

arukiidou:
rebased it.

arukiidou:
re-updates to v2.0.2

leogr:
cc @loresuso @alacuku

leogr (approved):

PR #230, opened 2023-01-24
fix(pusher): automatically add latest only if no other tag was specified

LucaGuerra:
Signed-off-by: Luca Guerra <luca@guerra.sh>

**What type of PR is this?**

/kind bug

/area library

When calling the `Push()` function, if tags are specified via `WithTags()` with a repo name without an explicit reference in the name, only the provided tags should be pushed and not `latest`. In all other cases, the behavior should not change.

Fixes #

### DIFF
--- a/pkg/oci/pusher/pusher.go
+++ b/pkg/oci/pusher/pusher.go
@@ -98,9 +98,15 @@ func (p *Pusher) Push(ctx context.Context, artifactType oci.ArtifactType,
 		return nil, err
 	}
 
+	tags := o.Tags
+
 	// Using ":latest" by default if no tag was provided.
 	if repo.Reference.Reference == "" {
-		repo.Reference.Reference = oci.DefaultTag
+		if len(tags) > 0 {
+			repo.Reference.Reference, tags = tags[0], tags[1:]
+		} else {
+			repo.Reference.Reference = oci.DefaultTag
+		}
 	}
 
 	// Set remoteTarget and its tracker.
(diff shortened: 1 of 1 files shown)


pkg/oci/pusher/pusher.go
```diff
@@ -172,10 +167,18 @@ func (p *Pusher) Push(ctx context.Context, artifactType oci.ArtifactType,
...
+	if repo.Reference.Reference == "" && len(o.Tags) == 0 {
+		repo.Reference.Reference = oci.DefaultTag
+	}
+
+	if repo.Reference.Reference != "" {
+		// Tag the root descriptor remotely.
+		err = repo.PushReference(ctx, *rootDesc, rootReader, repo.Reference.Reference)
+		if err != nil {
+			return nil, err
+		}
```
loresuso, line 181: What if you pass a ref without tag (repo.Reference.Reference == "") and len(o.Tags) != 0?
The subsequent TagN function at line 187 will fail because an empty ref is passed. We should find a way to handle this

LucaGuerra: oops, I'll take a closer look at `TagN` and update

LucaGuerra:
@loresuso thanks for the comment! Should be right now, please take another look :sweat_smile:

loresuso (approved):
@LucaGuerra yes, this is how I figured it in my mind too, thanks! 
/LGTM

FedeDP (approved):
/approve

PR #235, opened 2023-01-26
fix(internal): error handling with no artifacts

therealbobo:
Signed-off-by: Roberto Scolaro <roberto.scolaro21@gmail.com>

**What type of PR is this?**

> /kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

Using the `falcoctl artifact install` with no ref results in no error.

Fixes #

NONE

### DIFF
--- a/cmd/artifact/install/install.go
+++ b/cmd/artifact/install/install.go
@@ -91,12 +91,12 @@ func NewArtifactInstallCmd(ctx context.Context, opt *options.CommonOptions) *cob
 
 // RunArtifactInstall executes the business logic for the artifact install command.
 func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []string) error {
-	if len(args) == 0 {
-		configuredInstaller, err := config.Installer()
-		if err != nil {
-			o.Printer.CheckErr(fmt.Errorf("unable to retrieved the configured installer: %w", err))
-		}
+	configuredInstaller, err := config.Installer()
+	if err != nil {
+		o.Printer.CheckErr(fmt.Errorf("unable to retrieved the configured installer: %w", err))
+	}
 
+	if len(args) == 0 {
 		// Set args as configured if no arg was passed
 		if len(configuredInstaller.Artifacts) == 0 {
 			return fmt.Errorf("no artifacts to install, please configure artifacts or pass them as arguments to this command")
--- a/cmd/artifact/install/install.go
+++ b/cmd/artifact/install/install.go
@@ -97,6 +97,10 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 			o.Printer.CheckErr(fmt.Errorf("unable to retrieved the configured installer: %w", err))
 		}
 
+		// Set args as configured if no arg was passed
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -368,7 +368,9 @@ func Installer() (Install, error) {
 	// with Install we can just use nested keys.
 	// env variables can just make use of ";" to separat
 	artifacts := viper.GetStringSlice(ArtifactInstallArtifactsKey)
-	if len(artifacts) == 1 { // in this case it might come from the env
+	if len(artifacts) == 0 {
+		return Install{}, fmt.Errorf("cannot find any artifact to install")
+	} else if len(artifacts) == 1 { // in this case it might come from the env
```
loresuso, line 373: This is a good point, but not the right the right point to do it in the code in my opinion.
Here we are dealing with the configuration (config file + env variables), and you are saying that the install should fail if the user didn't configured any artifacts. However, artifacts can be passed as command line arguments to the install command. In that case, those are the artifacts to be used and install. So what I would do, is moving the check in the command itself, so that if neither the config or the args contain artifacts, the command fails. Does it make sense to you? Thank you :)

loresuso: See this for instance https://github.com/falcosecurity/falcoctl/blob/main/cmd/artifact/follow/follow.go#L208

therealbobo: You are completely right! Makes a lot of sense! Thank you for the feedback! :D

cmd/artifact/install/install.go
```diff
@@ -97,6 +97,10 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 			o.Printer.CheckErr(fmt.Errorf("unable to retrieved the configured installer: %w", err))
 		}
 
+		// Set args as configured if no arg was passed
+		if len(configuredInstaller.Artifacts) == 0 {
+			return fmt.Errorf("no artifacts to install, please configure artifacts or pass them as arguments to this command")
+		}
```
loresuso, line 103: This is ok for the args and refs of the config. However, there is another point to address: here we are considering the config only if `args` len is equal to zero. But this is not completely correct, since we want to load the config in any case because it may contain other parameters, like for instance the plugins dir and rulesfiles dir. So I would write something like this at the beginning of installer:
```
	// Retrieve configuration for installer
	configuredInstaller, err := config.Installer()
	if err != nil {
		o.Printer.CheckErr(fmt.Errorf("unable to retrieved the configured installer: %w", err))
	}

	// Set args as configured if no arg was passed
	if len(args) == 0 {
		if len(configuredInstaller.Artifacts) == 0 {
			return fmt.Errorf("no artifacts to install, please configure artifacts or pass them as arguments to this command")
		}
		args = configuredInstaller.Artifacts
	}
```

therealbobo: Aah! That's right! Thank you! :D

loresuso (approved):
Thanks a lot! 

/approve

alacuku (approved):
Thanks!

leogr (approved):

PR #244, opened 2023-01-31
new(index-caching): add in-memory cache mechanism for index files

alacuku:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

Since index files can be overwritten using ENV variables we needed a new solution to handle them on a per-command basis. Added an in-memory caching system that controls the index files during the command execution.
Furthermore, it handles loading and saving the index cache from/to the filesystem.
The files that track the indexes are saved to ~/.config/falcoctl.

Fixes #

### DIFF
--- a/cmd/artifact/artifact.go
+++ b/cmd/artifact/artifact.go
@@ -26,10 +26,10 @@ import (
 	"github.com/falcosecurity/falcoctl/cmd/artifact/install"
 	"github.com/falcosecurity/falcoctl/cmd/artifact/list"
 	"github.com/falcosecurity/falcoctl/cmd/artifact/search"
-	"github.com/falcosecurity/falcoctl/cmd/index/add"
 	"github.com/falcosecurity/falcoctl/cmd/registry/auth/basic"
 	"github.com/falcosecurity/falcoctl/cmd/registry/auth/oauth"
 	"github.com/falcosecurity/falcoctl/internal/config"
+	"github.com/falcosecurity/falcoctl/pkg/index/cache"
 	commonoptions "github.com/falcosecurity/falcoctl/pkg/options"
 )
 
@@ -49,13 +49,9 @@ func NewArtifactCmd(ctx context.Context, opt *commonoptions.CommonOptions) *cobr
 			indexes, err := config.Indexes()
 			opt.Printer.CheckErr(err)
 
-			for _, ind := range indexes {
-				indexMgr := add.IndexAddOptions{
-					CommonOptions: opt,
-				}
-				opt.Printer.CheckErr(indexMgr.Validate([]string{ind.Name, ind.URL}))
-				opt.Printer.CheckErr(indexMgr.RunIndexAdd(ctx, []string{ind.Name, ind.URL}))
-			}
+			// Create the index cache.
+			indexCache, err := cache.NewFromConfig(ctx, config.IndexesFile, config.IndexesDir, indexes)
+			opt.Printer.CheckErr(err)
 
 			basicAuths, err := config.BasicAuths()
 			opt.Printer.CheckErr(err)
@@ -81,6 +77,8 @@ func NewArtifactCmd(ctx context.Context, opt *commonoptions.CommonOptions) *cobr
 				}
 				opt.Printer.CheckErr(oauthMgr.RunOauth(ctx, []string{auth.Registry}))
 			}
+
+			opt.Initialize(commonoptions.WithIndexCache(indexCache))
 		},
 	}
 
--- a/cmd/artifact/follow/follow.go
+++ b/cmd/artifact/follow/follow.go
@@ -261,7 +261,7 @@ func (o *artifactFollowOptions) RunArtifactFollow(ctx context.Context, args []st
 		} else {
 			o.Printer.Info.Printfln("Creating follower for %q, with check every %s", a, o.every.String())
 		}
-		ref, err := utils.ParseReference(mergedIndexes, a)
+		ref, err := o.IndexCache.ResolveReference(a)
 		if err != nil {
 			return fmt.Errorf("unable to parse artifact reference for %q: %w", a, err)
 		}
--- a/cmd/index/add/add.go
+++ b/cmd/index/add/add.go
@@ -17,7 +17,6 @@ package add
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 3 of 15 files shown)


cmd/artifact/follow/follow.go
```diff
@@ -261,7 +261,7 @@ func (o *artifactFollowOptions) RunArtifactFollow(ctx context.Context, args []st
 		} else {
 			o.Printer.Info.Printfln("Creating follower for %q, with check every %s", a, o.every.String())
 		}
-		ref, err := utils.ParseReference(mergedIndexes, a)
+		ref, err := o.IndexCache.ParseReference(a)
```
leogr, line 264: Better than before, but it seems a bit weird to me that `ParseReference(a)` belongs to a _cache_ thing.
Not a blocker, anyway.

cmd/artifact/artifact.go
```diff
@@ -49,12 +49,13 @@ func NewArtifactCmd(ctx context.Context, opt *commonoptions.CommonOptions) *cobr
 			indexes, err := config.Indexes()
 			opt.Printer.CheckErr(err)
 
+			// Create the index cache.
+			indexCache, err := cache.New(config.IndexesFile, config.IndexesDir)
+			opt.Printer.CheckErr(err)
```
leogr, line 54: I like that :+1:

cmd/index/add/add.go
```diff
@@ -78,64 +76,33 @@ func NewIndexAddCmd(ctx context.Context, opt *options.CommonOptions) *cobra.Comm
...
 
-	indexConfig.Add(entry)
-
-	if err := indexConfig.Write(config.IndexesFile); err != nil {
-		return err
+	for _, cfg := range indexes.Configs {
+		currentIndexes = append(currentIndexes, config.Index{
+			Name: cfg.Name,
+			URL:  cfg.URL,
+		})
```
leogr, line 103: By definition, a cache is ephemeral. So I'd expect these two steps in reverse order:
- first, store the config (no matter what will happen with the cache)
- second, store the cache (if can't store the cache is not a real problem, it can be recreated at any time).

Note:
not really important for now

leogr (commented):
I really like these improvements :star_struck: 

Left some comments after a quick look. I reserve some time to look deeper into this.


Question regarding `func (i *Index) Write(path string) error` :point_down: 
https://github.com/falcosecurity/falcoctl/pull/244/files#diff-0c91f7868027b72ed711850b021be1633b12516a7a4b3c62515f27291e88fcf6L132-L147

Is there any specific reason to do not to use something like https://pkg.go.dev/os#MkdirAll?
In order words, why is the responsibility of creating any necessary parent dirs delegated to other code paths (and included in that func?)

cmd/artifact/follow/follow.go
alacuku: The *cache* thing here is and abstraction built upon the indexes/mergeIndexes. We need the *mergedIndexes* info in order to "calculate" which reference to use for an artifact. I agree that we should change the name of the function instead of moving it around.

alacuku:
> 



> Question regarding `func (i *Index) Write(path string) error` point_down https://github.com/falcosecurity/falcoctl/pull/244/files#diff-0c91f7868027b72ed711850b021be1633b12516a7a4b3c62515f27291e88fcf6L132-L147
> 
> Is there any specific reason to do not to use something like https://pkg.go.dev/os#MkdirAll? In order words, why is the responsibility of creating any necessary parent dirs delegated to other code paths (and included in that func?)

No particular reason, I could add it in a different PR.

cmd/index/add/add.go
alacuku: Totally agree about the cache,  still, I would prefer that the changes to the config file are done only if the run of the command is successful. The user should not expect anything if the command fails. 
Btw, I changed the logic of the config file update.

alacuku:
> > 
> 
> > Question regarding `func (i *Index) Write(path string) error` point_down https://github.com/falcosecurity/falcoctl/pull/244/files#diff-0c91f7868027b72ed711850b021be1633b12516a7a4b3c62515f27291e88fcf6L132-L147
> > Is there any specific reason to do not to use something like https://pkg.go.dev/os#MkdirAll? In order words, why is the responsibility of creating any necessary parent dirs delegated to other code paths (and included in that func?)
> 
> No particular reason, I could add it in a different PR.

Added it in a different commit.

pkg/index/cache/cache.go
```diff
@@ -0,0 +1,315 @@
...
+	// Track the new indexes that need to be saved locally when writing the cache to file.
+	fetchedIndexes []*index.Index
+	// Track the indexes that have been removed, needed when writing the cache to file.
+	removedIndexes []string
+}
+
+// New creates a new cache object. For each entry in the indexes.yaml file it loads the respective index file
+// found on the disk or fetches it if not found. If there is an entry in the indexes.yaml file but its index file does not exist on the disk
+// then it will error.
+func New(ctx context.Context, indexFile, indexesDir string) (*Cache, error) {
```
leogr, line 46: Although it looks good to me, the documentation does not describe what an `indexFile` is.
In this context, I assume an _`indexFile` is the index file of the artifact's indexes files contained into `indexesDir`_. Is that correct?

pkg/index/index.go
```diff
@@ -222,6 +237,69 @@ func (m *MergedIndexes) IndexByEntry(entry *Entry) *Index {
...
+//     if instead a tag or a digest is specified, the name will be used to look up
+//     into mergedIndexes, then the tag or digest will be appended.
+//     e.g "cloudtrail:0.5.1" -> "ghcr.io/falcosecurity/plugins/cloudtrail:0.5.1"
+//     e.g "cloudtrail@sha256:123abc..." -> "ghcr.io/falcosecurity/plugins/cloudtrail@sha256:123abc...
+//
+//  2. if name is a reference without tag or digest, tag latest is appended.
+//     e.g. "ghcr.io/falcosecurity/plugins/cloudtrail" -> "ghcr.io/falcosecurity/plugins/cloudtrail:latest"
+//
+//  3. if name is a complete reference, it will be returned as is.
+func (m *MergedIndexes) ParseReference(name string) (string, error) {
```
leogr, line 254:
```suggestion
func (m *MergedIndexes) ResolveReference(name string) (string, error) {
```

This function is probably misnamed. It seems that it takes a relative name or a partial reference and resolves it to a fully qualified artifact reference. Thus my renaming suggestion.

leogr (commented):
SGTM now!

Just some nitpicking, see my comments below :clap:

internal/config/config.go
```diff
@@ -403,3 +403,56 @@ func UpdateConfigFile(key string, value interface{}, path string) error {
 // that we have fields like engine_version that are only numbers, we shoud be
 // as muche generic as possible.
 type FalcoVersions map[string]string
+
```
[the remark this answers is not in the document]
alacuku: It is the default value for the *--every* flag in the `artifact follow` command: https://github.com/falcosecurity/falcoctl/blob/38439929f12b012011647fb1c97550629967256a/cmd/artifact/follow/follow.go#L195-L196

pkg/index/cache/cache.go
alacuku: That's correrct. We did not document it, yet. Will do in the future, since it is an internal configuration and not a user facing one.

pkg/index/index.go
loresuso: Makes sense!

leogr (approved):

PR #247, opened 2023-02-01
fix: create artifact destination dir if not present

therealbobo:
Signed-off-by: Roberto Scolaro <roberto.scolaro21@gmail.com>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /area library

> /area cli

> /area tests

> /area examples

If the destination directory for the artifacts (`--rulesfiles-dir` or `--plugins-dir`) doesn't exist, the command fails.

Fixes #

NONE

### DIFF
--- a/cmd/artifact/follow/follow.go
+++ b/cmd/artifact/follow/follow.go
@@ -151,6 +151,10 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 					o.Printer.CheckErr(fmt.Errorf("unable to overwrite \"rulesfiles-dir\" flag: %w", err))
 				}
 			}
+			// Check if directory exists and is writable
+			if err := utils.ExistsAndIsWritable(f.Value.String()); err != nil {
+				o.Printer.CheckErr(fmt.Errorf("rulesfiles-dir: %w", err))
+			}
 
 			// Override "plugins-dir" flag with viper config if not set by user.
 			f = cmd.Flags().Lookup("plugins-dir")
@@ -162,6 +166,10 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 				if err := cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val)); err != nil {
 					o.Printer.CheckErr(fmt.Errorf("unable to overwrite \"plugins-dir\" flag: %w", err))
 				}
+			}
+			// Check if directory exists and is writable
+			if err := utils.ExistsAndIsWritable(f.Value.String()); err != nil {
+				o.Printer.CheckErr(fmt.Errorf("plugins-dir: %w", err))
 			}
 
 			// Override "tmp-dir" flag with viper config if not set by user.
--- a/cmd/artifact/install/install.go
+++ b/cmd/artifact/install/install.go
@@ -239,6 +239,13 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 			destDir = o.rulesfilesDir
 		}
 
+		if _, err = os.Stat(destDir); os.IsNotExist(err) {
+			err = os.MkdirAll(destDir, 0o700)
+			if err != nil {
+				return err
+			}
+		}
+
 		sp, _ := o.Printer.Spinner.Start(fmt.Sprintf("INFO: Extracting and installing %q %q", result.Type, result.Filename))
 		result.Filename = filepath.Join(tmpDir, result.Filename)
 
--- a/cmd/artifact/install/install.go
+++ b/cmd/artifact/install/install.go
@@ -97,6 +97,10 @@ func NewArtifactInstallCmd(ctx context.Context, opt *options.CommonOptions) *cob
 					o.Printer.CheckErr(fmt.Errorf("unable to overwrite \"rulesfiles-dir\" flag: %w", err))
 				}
 			}
+			// Check if directory exists and is writable
+			if err := utils.ExistsAndIsWritable(f.Value.String()); err != nil {
+				o.Printer.CheckErr(fmt.Errorf("rulesfiles-dir: %w", err))
+			}
 
 			// Override "plugins-dir" flag with viper config if not set by user.
 			f = cmd.Flags().Lookup("plugins-dir")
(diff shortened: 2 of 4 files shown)


cmd/artifact/install/install.go
```diff
@@ -239,6 +239,13 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 			destDir = o.rulesfilesDir
 		}
 
+		if _, err = os.Stat(destDir); os.IsNotExist(err) {
+			err = os.MkdirAll(destDir, 0o700)
```
leogr, line 243: The installer (`falcoctl`) and the consumer (`falco`) are two different processes, with different permissions and capabilities. Moreover, the consumer can run into different situations.
Think the syscall use case: Falco runs as root. However, in the plugin's use cases, it may run as non-root.

Considering that, I'm a bit concerned that `0x700` may break some cases.

internal/follower/follower.go
```diff
@@ -188,6 +188,14 @@ func (f *Follower) follow(ctx context.Context) {
 
 	dstDir := f.destinationDir(res)
 
+	if _, err := os.Stat(dstDir); os.IsNotExist(err) {
+		err = os.MkdirAll(dstDir, 0o700)
```
leogr, line 192: Same as above

cmd/artifact/follow/follow.go
```diff
@@ -151,6 +153,18 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 					o.Printer.CheckErr(fmt.Errorf("unable to overwrite \"rulesfiles-dir\" flag: %w", err))
 				}
 			}
+			// Check if directory exists and is writable
+			rulesfilesDir := fmt.Sprintf("%v",f.Value)
```
loresuso, line 157: I think here we should get the value from viper. Otherwise, when the rulesfiles dir comes from config or env we miss it. 

```
viper.Get(config.ArtifactFollowRulesfilesDirKey)
```

cmd/artifact/follow/follow.go
```diff
@@ -151,6 +153,18 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
...
+			if _, err := os.Stat(rulesfilesDir); os.IsNotExist(err) {
+				o.Printer.CheckErr(fmt.Errorf("unable to find \"rulesfiles-dir\": %w", err))
+			} else {
+				if f, err := os.Create(filepath.Join(rulesfilesDir, "._check_writable")); err == nil {
+					f.Close()
+					os.Remove(f.Name())
+				} else {
+					o.Printer.CheckErr(fmt.Errorf("\"rulesfiles-dir\" is not writable"))
+				}
+			}
```
loresuso, line 167: Let's convert this into a `utils` function so that we do not duplicate code

cmd/artifact/follow/follow.go
```diff
@@ -163,6 +177,18 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 					o.Printer.CheckErr(fmt.Errorf("unable to overwrite \"plugins-dir\" flag: %w", err))
 				}
 			}
+			// Check if directory exists and is writable
+			pluginsDir := fmt.Sprintf("%v",f.Value)
```
loresuso, line 181: Same as above on taking it from viper

cmd/artifact/install/install.go
```diff
@@ -239,6 +239,13 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
 			destDir = o.rulesfilesDir
 		}
 
+		if _, err = os.Stat(destDir); os.IsNotExist(err) {
```
loresuso, line 242: this should be no more needed :)

therealbobo:
I reviewed the logic: now, only at the start, `falcoctl` checks if the artifact destination directories are present and writable and, if not, will exit with an error. On the contrary if, during the execution, the directories are removed or became read-only, it will only log an error. All of this to avoid wrong configurations at startup.

cmd/artifact/follow/follow.go
```diff
@@ -151,6 +153,18 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 					o.Printer.CheckErr(fmt.Errorf("unable to overwrite \"rulesfiles-dir\" flag: %w", err))
 				}
 			}
+			// Check if directory exists and is writable
+			rulesfilesDir := fmt.Sprintf("%v",f.Value)
+			if _, err := os.Stat(rulesfilesDir); os.IsNotExist(err) {
+				o.Printer.CheckErr(fmt.Errorf("unable to find \"rulesfiles-dir\": %w", err))
+			} else {
+				if f, err := os.Create(filepath.Join(rulesfilesDir, "._check_writable")); err == nil {
```
leogr, line 161: What if the `._check_writable` file already exits? :smile:

loresuso (approved):
/approve

loresuso:
cc @leogr

leogr:
@loresuso @therealbobo as discussed privately, my only concern is `._check_writable` :)
Otherwise, SGTM!

internal/utils/fs.go
```diff
@@ -38,3 +39,30 @@ func Move(oldPath, newPath string) error {
 
 	return nil
 }
+
+// ExistsAndIsWritable checks if the directory specified by the path exists and is writable.
+func ExistsAndIsWritable(path string) error {
+	info, err := os.Stat(path)
+	if errors.Is(err, os.ErrNotExist) {
```
alacuku, line 46:
```suggestion
	if errors.Is(err, errors.Is(err, fs.ErrNotExist)) {
```
We should use the `errors` package.

alacuku (approved):
Left a suggestion.

Furthermore, should we change the title of the PR since we are not creating the directory?

internal/utils/fs.go
alacuku: Ignore this suggestion, it's already doing what I meant. My bad!

leogr (approved):

PR #251, opened 2023-02-03
new: introduce allowedTypes

loresuso:
Signed-off-by: Lorenzo Susini <susinilorenzo1@gmail.com>

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

Falcoctl is able to download any kind of artifacts, including shared libraries that could be potentially used by Falco and loaded. For security reasons, we want to let users be able to decide which kind of artifacts falcoctl is allowed to pull by means of a whitelist. If the list is empty everything is allowed for ease of use. 

Fixes #

### DIFF
--- a/cmd/artifact/install/install.go
+++ b/cmd/artifact/install/install.go
@@ -71,6 +71,7 @@ type artifactInstallOptions struct {
 	*options.RegistryOptions
 	rulesfilesDir string
 	pluginsDir    string
+	allowedTypes  oci.ArtifactTypeSlice
 }
 
 // NewArtifactInstallCmd returns the artifact install command.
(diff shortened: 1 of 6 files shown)


cmd/artifact/install/install.go
```diff
@@ -128,6 +144,8 @@ func NewArtifactInstallCmd(ctx context.Context, opt *options.CommonOptions) *cob
 		"directory where to install rules. Defaults to /etc/falco")
 	cmd.Flags().StringVarP(&o.pluginsDir, "plugins-dir", "", config.PluginsDir,
 		"directory where to install plugins. Defaults to /usr/share/falco/plugins")
+	cmd.Flags().Var(&o.allowedTypes, "allowed-types",
+		"whitelist of artifacts type that can be installed. If an empty list is passed (or configured), everything is allowed")
```
LucaGuerra, line 148:
```suggestion
		"comma-separated list of artifact types that can be installed. If not specified, all types are allowed. Example: \"rulesfile,plugin\"")
```

LucaGuerra: I think this phrasing is a little clearer but the flag name and everything as implemented makes sense to me :+1:

loresuso: Agreed but it isn't only a comma-separated list, you can also specify the flag multiple times. Any other hint?

LucaGuerra: Thanks @loresuso , I think your solution is the best since it explains all ways to use the flag!

loresuso: 👍 updated!

leogr (approved):
:+1:

PR #259, opened 2023-02-22
fix(artifact/info): print tags in their original format

alacuku:
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>

**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

This PR changes the output of the `artifact info` command. Available `tags` are now printed as retrieved from the remote repository.

Old output:
```bash
REF                                     TAGS                              
ghcr.io/falcosecurity/rules/falco-rules 0.0.0 0.1.0-rc1 0.1.0 0.1.0 latest
```
New output:
```bash
REF                                     TAGS                            
ghcr.io/falcosecurity/rules/falco-rules latest, 0.1.0, 0.1, 0, 0.1.0-rc1
```

Fixes #

### DIFF
--- a/pkg/oci/repository/repository.go
+++ b/pkg/oci/repository/repository.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/blang/semver"
 	"oras.land/oras-go/v2/registry/remote"
 )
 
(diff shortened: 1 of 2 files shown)


alacuku:
Hey @loresuso, could you have a look, please?

leogr (approved):

pkg/oci/repository/repository.go
```diff
@@ -73,41 +72,5 @@ func (r *Repository) Tags(ctx context.Context) ([]string, error) {
...
 
-	result, err = sortTags(result)
-	if err != nil {
-		return nil, err
-	}
-
-	return result, nil
-}
-
-func sortTags(tags []string) ([]string, error) {
```
loresuso, line 84: I think I would preserve at least a lexicographic order here!

alacuku: I think we should print them based on the order coming from the repository. From the most recent to the oldest one.

loresuso: Are we sure every registry will retrieve tags in that order? And even if so, imagine later in time we put a patch release of something so that you could have
latest, 0.1.0, 0.1, 0, 0.1.0-rc1, 1.0.0, 0.1.1
We can end up with mixed versions, as a user I would prefer it to be ordered some way, wdyt?

alacuku: The OCI spec says that registries must return the tags in lexical order: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-tags

loresuso: Alright, perfect then!

alacuku: It seems that `ghcr.io` does not comply with the spec. It returns them based on the creating order: most recent to oldest. Anyway, in the future, we could force the tags to be ordered if any use case pops out.

loresuso: Makes sense

PR #278, opened 2023-04-26
fix(registry/push): create tar.gz archive for raw input files

alacuku:
**What type of PR is this?**

/kind bug

/kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

/area library

/area cli

/area tests

> /area examples

When pushing a new artifact to a remote registry it checks the input file and if not in tar.gz format it compresses it and archives it in tar format. The `mediaType` for our artifacts is in the `tar.gz` format, so we need to push them in that format.

Furthermore, error messages are handled in the root command. All subcommands propagate the fatal errors to the root command. This simplifies the error
handling:
  * commands do not exit in the middle of their business logic if a fatal error occurs, just return the error;
  * when an error occurs the application exits with error code != 0, allowing consumers to know that something went wrong without the need to check the output messages;
  * output is consistent among subcommands, in case of error the usage info is printed out(only not leaf subcommands) and the error message is formatted using the printer;
  * easies the testing of the application.

Integration tests have been written for the `falcoctl registry push` command, that's the reason why this PR has so many added lines.

Fixes #

The changes are contained in their own commits. I preferred to open a single PR to group them together since they are all related.


cmd/registry/push/push.go
```diff
@@ -127,6 +147,30 @@ func (o *pushOptions) RunPush(ctx context.Context, args []string) error {
...
+			return err
+		} else if err == nil {
+			continue
+		} else {
+			path, err := utils.CreateTarGzArchive(p)
+			if err != nil {
+				return err
+			}
+			paths[i] = path
+			if toBeDeleted == "" {
```
FedeDP, line 168: What happens if multiple `paths` need to be archived? Are we going to cleanup only the first one?

FedeDP: Uh i see, we store the folder, not the filepath. It's ok then, sorry!

internal/utils/compress.go
```diff
@@ -0,0 +1,94 @@
...
+			}
+		}
+	}()
+
+	fInfo, err := os.Stat(cleanedPath)
+	if err != nil {
+		return "", err
+	}
+
+	header, err := tar.FileInfoHeader(fInfo, fInfo.Name())
```
FedeDP, line 73: I'd move this check before opening the writers
!

internal/utils/compress.go
```diff
@@ -0,0 +1,94 @@
...
+	outFile, err := os.Create(filepath.Clean(filepath.Join(tmpDir, nameTokens[0]+".tar.gz")))
+	if err != nil {
+		return "", err
+	}
+
+	// Create new writer for gzip.
+	gzw := gzip.NewWriter(outFile)
+	defer func() {
+		if err == nil {
+			err = gzw.Close()
```
FedeDP, line 49: Note: we are not propagating the `err` to the caller here; if you want to do so, we need to use a named return value, ie:
```
func CreateTarGzArchive(path string) (myRes string, err error) {
```

See this snippet: https://go.dev/play/p/Gqcc7JlIDx3

internal/utils/compress.go
```diff
@@ -0,0 +1,94 @@
...
+		} else {
+			if errDefer := gzw.Close(); errDefer != nil {
+				err = fmt.Errorf("%s: %w", err.Error(), errDefer)
+			}
+		}
+	}()
+
+	tw := tar.NewWriter(gzw)
+	defer func() {
+		if err == nil {
```
FedeDP, line 59: Same as above; i don't know if this pattern is used elsewhere in the project too!

internal/utils/compress.go
alacuku: Nice catch!

internal/utils/fileTypes.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return fmt.Errorf("%s: %w", err.Error(), ErrNotTarGz)
+	}
+
+	tarReader := tar.NewReader(uncompressedStream)
+
+	// Loop through the files and check if the header is ok.
+	// If the files are not in tar format it will error.
+	for {
+		_, err = tarReader.Next()
```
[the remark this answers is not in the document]
alacuku: If there are multiple files we need to check them all.

cmd/artifact/artifact.go
```diff
@@ -40,45 +37,47 @@ func NewArtifactCmd(ctx context.Context, opt *commonoptions.CommonOptions) *cobr
 		DisableFlagsInUseLine: true,
 		Short:                 "Interact with Falco artifacts",
 		Long:                  "Interact with Falco artifacts",
-		PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		SilenceErrors:         true,
+		SilenceUsage:          true,
```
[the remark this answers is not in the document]
alacuku: The idea here is that the leaf commands do not print the usage info. They return the error, and the error is printed out by the root command. The user has to run the command with `--help` flag to get the usage info. For all the non leaf commands the usage is printed out automatically.

FedeDP:
This LGTM! If CI becomes :green_circle:  i'll approve :D

FedeDP (approved):
Great job!
/approve

PR #280, opened 2023-05-04
fix: check directory existance iff necessary

### LINKED ISSUE #253: falcoctl install/follow fails if "plugins-dir" and/or "rulesfiles-dir" does not exist
**What happened**:
When installing a plugin artifact `falcoctl` fails because the `rulesfiles-dir` does not exist:
```bash
falcoctl artifact install k8saudit --plugins-dir ./
ERRO  rulesfiles-dir: /etc/falco doesn't exists
```

The command should fail only if the `plugins-dir` does not exist or is not writable by the tool. In this case, the command should run successfully and not fail because the `rulesfiles-dir` is not needed.

Make sure that the default(/etc/falco) "rulesfiles-dir" does not exist and run the following command:
```bash
falcoctl artifact install k8saudit --plugins-dir ./
```

The current behavior was introduced in the following PR: #247.

therealbobo:
**What type of PR is this?**

/kind bug

/area cli

The directory existence and possibility to write is checked only if necessary.

Fixes #253 

### DIFF
--- a/cmd/artifact/follow/follow.go
+++ b/cmd/artifact/follow/follow.go
@@ -156,10 +156,6 @@ func NewArtifactFollowCmd(ctx context.Context, opt *options.CommonOptions) *cobr
 					return fmt.Errorf("unable to overwrite %q flag: %w", install.FlagRulesFilesDir, err)
 				}
 			}
-			// Check if directory exists and is writable
-			if err := utils.ExistsAndIsWritable(f.Value.String()); err != nil {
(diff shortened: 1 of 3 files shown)


maxgio92:
/lgtm

WDYT @alacuku?

loresuso (approved):
LGTM /approve

alacuku (approved):
Thanks @therealbobo!

/LGTM

leogr (approved):

PR #288, opened 2023-05-22
feat(cmd/auth) New command to support Workload Identity on GKE

### LINKED ISSUE #277: Support Workload Identity for oauth login
Support of [workload identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity) to retrieve an access token and connect to a private Artifact registry, using the `falcoctl registry auth oauth` command.

**Why is this needed**:
To take advantage of workload authentication and avoid using secrets in a GKE environment.

MickaelFontes:
To add support of workload identity to authenticate to a Artifact Registry.

In this PR, I chose to add a `auth gcp` command, which is a mix of both existing `basic` and `oauth` commands.
Some context:

* to authenticate to Artifact Registry, you have to use a short or long-term token as a password
* to avoid to use long term tokens (service account keys), GCP allow you to get your pod  inside a GKE cluster authenticated to retrieve a short-term oauth2 access token (see [Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#verify_the_setup))
* this is not possible using the `auth oauth` command (with uses the [`oauth2/clientcredentials`](https://github.com/golang/oauth2/tree/master/clientcredentials) but it can be possible using the [`oauth2/google`](https://github.com/golang/oauth2/blob/master/google/google.go) : this is the goal of this PR.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

Fixes #277 

* I have not extensively tested it for now (only a few pull), it's the beginning.
* I edited the `go.mod` file not on purpose when adding the imports :grimacing: 
* I had to make a choice on how to add Workload Identity support for Artifact Registry and decided to go for the `auth gcp` command, but I don't know if you if it is a good choice or not.
* It's my first Go project (hence the previous point I think :sweat_smile:) so guidance/help and review is appreciated :smile:

### DIFF
--- a/cmd/registry/auth/gcp/gcp.go
+++ b/cmd/registry/auth/gcp/gcp.go
@@ -26,7 +26,9 @@ import (
 )
 
 const (
-	longGcp = `Register a registry to use GCE Metadata server or gcloud Application Default credentials to connect to it.
+	longGcp = `Register an Artifact Registry to use GCP Application Default credentials to connect to it.
+
+In particular, it can use Workload Identity or GCE metadata server to authenticate.
 
 Example 
 	falcoctl registry auth gcp europe-docker.pkg.dev
@@ -47,7 +49,7 @@ func NewGcpCmd(ctx context.Context, opt *options.CommonOptions) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:                   "gcp [REGISTRY]",
 		DisableFlagsInUseLine: true,
-		Short:                 "Register an OCI registry to log in using GCP common credentials",
+		Short:                 "Register an Artifact Registry to log in using GCP Application Default credentials",
 		Long:                  longGcp,
 		Args:                  cobra.ExactArgs(1),
 		SilenceErrors:         true,
--- a/cmd/registry/auth/gcp/gcp.go
+++ b/cmd/registry/auth/gcp/gcp.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package gke
+package gcp
 
 import (
 	"context"
@@ -21,63 +21,63 @@ import (
 	"github.com/spf13/cobra"
 
 	"github.com/falcosecurity/falcoctl/internal/config"
-	"github.com/falcosecurity/falcoctl/internal/login/gke"
+	"github.com/falcosecurity/falcoctl/internal/login/gcp"
 	"github.com/falcosecurity/falcoctl/pkg/options"
 )
 
 const (
-	longGke = `Register a registry to use Workload Identity to connect to it.
+	longGcp = `Register a registry to use GCE Metadata server or gcloud Application Default credentials to connect to it.
 
 Example 
-	falcoctl registry gke europe-docker.pkg.dev
+	falcoctl registry gcp europe-docker.pkg.dev
 `
 )
 
-// RegistryGkeOptions contains the options for the registry gke command.
-type RegistryGkeOptions struct {
+// RegistryGcpOptions contains the options for the registry gcp command.
+type RegistryGcpOptions struct {
 	*options.CommonOptions
 	registry string
 }
 
-// NewGkeCmd returns the gke command.
-func NewGkeCmd(ctx context.Context, opt *options.CommonOptions) *cobra.Command {
-	o := RegistryGkeOptions{
+// NewGcpCmd returns the gcp command.
+func NewGcpCmd(ctx context.Context, opt *options.CommonOptions) *cobra.Command {
+	o := RegistryGcpOptions{
 		CommonOptions: opt,
 	}
 
 	cmd := &cobra.Command{
-		Use:                   "gke [REGISTRY]",
+		Use:                   "gcp [REGISTRY]",
 		DisableFlagsInUseLine: true,
-		Short:                 "Register an OCI registry to log in using Workload identity",
-		Long:                  longGke,
+		Short:                 "Register an OCI registry to log in using GCP common credentials",
+		Long:                  longGcp,
 		Args:                  cobra.ExactArgs(1),
 		SilenceErrors:         true,
 		SilenceUsage:          true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			return o.RunGke(ctx, args)
+			return o.RunGcp(ctx, args)
 		},
 	}
 
 	return cmd
 }
 
-// RunGke executes the business logic for the gke command.
-func (o *RegistryGkeOptions) RunGke(ctx context.Context, args []string) error {
+// RunGcp executes the business logic for the gcp command.
+func (o *RegistryGcpOptions) RunGcp(ctx context.Context, args []string) error {
 	var err error
 	reg := args[0]
-	if err = gke.Login(ctx, reg); err != nil {
+	if err = gcp.Login(ctx, reg); err != nil {
 		return err
 	}
-	o.Printer.Success.Printfln("GKE source correctly set for %q", o.registry)
+	o.Printer.Success.Printfln("GCP authentication successful for %q", reg)
 
-	o.Printer.Verbosef("Adding new gke entry to configuration file %q", o.ConfigFile)
-	if err = config.AddGke([]config.GkeAuth{{
+	o.Printer.Verbosef("Adding new gcp entry to configuration file %q", o.ConfigFile)
+	if err = config.AddGcp([]config.GcpAuth{{
 		Registry: reg,
 	}}, o.ConfigFile); err != nil {
 		return fmt.Errorf("index entry %q: %w", reg, err)
 	}
 
-	o.Printer.Success.Printfln("Gke auth entry for %q successfully added", reg)
+	o.Printer.Success.Printfln("GCP authentication entry for %q successfully added in configuration file", reg)
 
 	return nil
 }
--- a/cmd/registry/auth/gcp/gcp.go
+++ b/cmd/registry/auth/gcp/gcp.go
@@ -36,7 +36,6 @@ Example
 // RegistryGcpOptions contains the options for the registry gcp command.
 type RegistryGcpOptions struct {
 	*options.CommonOptions
-	registry string
 }
 
 // NewGcpCmd returns the gcp command.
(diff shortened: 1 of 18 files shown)


leogr:
> Thanks for your pull request. Before we can look at it, you'll need to add a 'DCO signoff' to your commits.
> 
> memo **Please follow instructions in the [contributing guide](https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md) to update your commits with the DCO**
> 
> Full details of the Developer Certificate of Origin can be found at [developercertificate.org](https://developercertificate.org/).
> 
> **The list of commits missing DCO signoff**:
> 
> * [eaa6086](https://github.com/falcosecurity/falcoctl/commits/eaa6086923fe4f07777214d5495293adae7b89aa) feat(registry/auth): Beginning of Workload Identity for falcoctl

Hey @MickaelFontes 

Thank you for this PR! Could you [sign off](https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md#developer-certificate-of-origin) your commits, please?
It is required by our policy :pray:

MickaelFontes:
Woops sorry for the forgotten sign off @leogr , it is done now!

Just for the record, this PR might not move for the coming days, since a refactor is needed before the actual add of workload identity support.

pkg/oci/authn/gke.go
```diff
@@ -0,0 +1,76 @@
...
+	// if we did not already load a token source for this registry check the config file
+	if !exists {
+		gkeAuths, err := config.Gkes()
+		if err != nil {
+			return auth.EmptyCredential, fmt.Errorf("unable to retrieve gke credentials %w", err)
+		}
+
+		for _, gke := range gkeAuths {
+			if gke.Registry == reg {
+				tokenSource, err = google.DefaultTokenSource(ctx)
```
max-frank, line 52: Nice work with the PR!

Might be premature since its still WIP, but when looking over it now I was wondering if we actually need one token source per registry here. Since unless I am wrong with workload identity we get only one identity per exec env anyway. Unless we add support for specifying different credential files per registry. So simply caching that workload identity is configured for the registry and then re-using the same token source for all should work. 🤔 💭

MickaelFontes: Hi max! Thank you very much for your review!
Indeed, I definitely knew some things are still to be disccused/not in their best form for now, and you pointed the finger at one (or two) of them :sweat_smile: 

First, I agree with you, no need to cache the Workload identity source, since it won't change for the same workload :sweat_smile: .
But here, I still used the `google.DefaultTokenSource(ctx)` instead of the `google.ComputeTokenSource("")` which only uses the GCE metadata server (or Workload Identity when that's the case). And this `DefaultTokenSource` (see [here](https://github.com/golang/oauth2/blob/master/google/default.go#L109) can use other sources when available (`GOOGLE_APPLICATION_CREDENTIALS` for instance).
Hence why I first saved the found tokenSource. But even right now I still doubt it's a good idea^^

MickaelFontes:
Hi all!
I pushed a first working version for Workload Identity support, and it opened some questions, for which any help is appreciated!
I'll list some points bothering me right now:

* using [`oauth/google`](https://github.com/golang/oauth2/tree/master/google), we can easily get a TokenSource that uses Workload Identity but we can get TokenSources using Application Default credentials or else (see [here](https://github.com/golang/oauth2/blob/master/google/default.go#L109)).
Hence my question about whether I should only use the Workload Identity Token source directly, or let the `google.DefaultTokenSource` choose among the available sources :thinking: 
(I started with the second approach initially, but the more I think about it, the more I tend to think it was a bad idea - and credential helper support with the refactor done by max makes Application Default credentials less relevant I think)
* As @max-frank  highlighted, caching the token source would be irrelevant if the only source is Workload Identity/GCE metadata, but what about if it is not and Application Default credentials are supported?
* one point I wanted to mention: the oauth2/google library offers the possibility to get the authenticated HTTP client by simply giving a TokenSource as input (see [here](https://github.com/golang/oauth2/blob/master/oauth2.go#L346) - I used that in my [PoC](https://github.com/falcosecurity/falcoctl/commit/c8a9c63c8fc2ca76ad52831396c676cb2be73fca))
To match the current Client creation logic, I instead used the TokenSource, assumed to get an access token and used it to perform a login/password authentication, as specified in Google's documentation [here](https://cloud.google.com/artifact-registry/docs/docker/authentication#token). Does this seems ok for you?
* last point more about a simple technical question: here I pushed new go.mod/go.sum files since I added a new library, but I don't think I used the proper way to add them :sweat_smile:. Could anyone tell me if the current state seems rights to him, and show me the proper way to edit them please :pray: ?

pkg/oci/authn/gke.go
max-frank: Ah no what I mean here is that we only need one DefaulTokenSource for all the registries. So we only need to load it once and then can re-use it for any registry that's set to workload identity/application default credentials

max-frank:
> * using [`oauth/google`](https://github.com/golang/oauth2/tree/master/google), we can easily get a TokenSource that uses Workload Identity but we can get TokenSources using Application Default credentials or else (see [here](https://github.com/golang/oauth2/blob/master/google/default.go#L109)).
>   Hence my question about whether I should only use the Workload Identity Token source directly, or let the `google.DefaultTokenSource` choose among the available sources 🤔

I think it makes the most sense to use `DefaultTokenSource` since thats expected behavior for most libraries supporting workload identity. 

>   (I started with the second approach initially, but the more I think about it, the more I tend to think it was a bad idea - and credential helper support with the refactor done by max makes Application Default credentials less relevant I think)
> * As @max-frank  highlighted, caching the token source would be irrelevant if the only source is Workload Identity/GCE metadata, but what about if it is not and Application Default credentials are supported?
With app default credentials (i.e., what. `DefaultTokenSource` resolves) we also only need 1 single `DefaultTokenSource` instance for the whole client (since it will always resolve to the same identity). So we can just load the `DefaultTokenSource` once and then re-use the same instance for all registries using GKE auth.

> * one point I wanted to mention: the oauth2/google library offers the possibility to get the authenticated HTTP client by simply giving a TokenSource as input (see [here](https://github.com/golang/oauth2/blob/master/oauth2.go#L346) - I used that in my [PoC](https://github.com/falcosecurity/falcoctl/commit/c8a9c63c8fc2ca76ad52831396c676cb2be73fca))
>   To match the current Client creation logic, I instead used the TokenSource, assumed to get an access token and used it to perform a login/password authentication, as specified in Google's documentation [here](https://cloud.google.com/artifact-registry/docs/docker/authentication#token). Does this seems ok for you?

Thats exactly the type of application logic I was expecting. This is also exactly how google implemented the logic in their own go-containerregistry lib (see https://github.com/google/go-containerregistry/blob/v0.15.2/pkg/v1/google/auth.go#L49-L61). If we want to add more functionality we could also add support for gcloud as configurable source for dev environments (though you can always run `gcloud auth application-default login` for dev anyway).

> * last point more about a simple technical question: here I pushed new go.mod/go.sum files since I added a new library, but I don't think I used the proper way to add them 😅. Could anyone tell me if the current state seems rights to him, and show me the proper way to edit them please 🙏 ?

It looks fine to me as long as you added it with
`go get <package>` and then after ran `go mod tidy` (for cleaning the dependency tree) it should be fine. You can also run `make fmt` and `make lint` to make sure everything is as the CI/CD pipeline will expect and accepts

alacuku:
Hi @MickaelFontes, could you please look at the linting issue?

MickaelFontes:
Hi @alacuku it's done! I think everything should be good now.
I've been testing on my side the new command for authentification, no issue found.

What should I do next to ensure everything is correct?

internal/config/config.go
```diff
@@ -393,6 +411,45 @@ func oathAuthListHookFunc() mapstructure.DecodeHookFuncType {
 	}
 }
 
+// oauthAuthListHookFunc returns a DecodeHookFunc that converts
+// strings to string slices, when the target type is DotSeparatedStringList.
+// when passed as env should be in the following format:
+// "registry;registry1".
+func gcpAuthListHookFunc() mapstructure.DecodeHookFuncType {
```
alacuku, line 418: We need to add the new ENV variable to the README.md: https://github.com/falcosecurity/falcoctl#falcoctl-environment-variables.

cmd/registry/auth/gcp/gcp.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/falcosecurity/falcoctl/internal/config"
+	"github.com/falcosecurity/falcoctl/internal/login/gcp"
+	"github.com/falcosecurity/falcoctl/pkg/options"
+)
+
+const (
+	longGcp = `Register a registry to use GCE Metadata server or gcloud Application Default credentials to connect to it.
+
+Example 
+	falcoctl registry gcp europe-docker.pkg.dev
```
alacuku, line 32:
```suggestion
	falcoctl registry auth gcp europe-docker.pkg.dev
```

cmd/registry/auth/gcp/gcp.go
```diff
@@ -0,0 +1,82 @@
...
+`
+)
+
+// RegistryGcpOptions contains the options for the registry gcp command.
+type RegistryGcpOptions struct {
+	*options.CommonOptions
+}
+
+// NewGcpCmd returns the gcp command.
+func NewGcpCmd(ctx context.Context, opt *options.CommonOptions) *cobra.Command {
```
alacuku, line 42: We need to add the new command to the README.md: https://github.com/falcosecurity/falcoctl#falcoctl-registry.
Feel free to add examples too.

alacuku:
Could you please rephrase the commit messages following this convention: https://github.com/falcosecurity/.github/blob/main/CONTRIBUTING.md#commit-convention?

MickaelFontes:
Thanks for the feedback I'll work on it!
About the commit messages, I thought that it would be okay if the PR is "squashed and merged" and the squashed commit follows the guidelines, right @alacuku ?
Otherwise of course I'll update the messages accordingly.

alacuku:
> About the commit messages, I thought that it would be okay if the PR is "squashed and merged" and the squashed commit follows the guidelines, right @alacuku ?

Yeah, squashing the commits it's fine! Thanks!

alacuku (approved):
/LGTM

FedeDP (approved):
/approve

alacuku:
Unfortunately, we can do nothing. I thought the commits were already squashed, my bad that I did not check.

Issue #300, opened 2023-06-06
Add support for more index hosting backends (e.g., GCS, S3, etc.)


max-frank:
**What would you like to be added**:  

Support for retrieving index files from storage backends using protocols other than plain HTTP/S. For example GCS or S3 storage for cloud native environments. Additionally support for authentication mechanisms for HTTP/S backends would also be helpful for private rule sets/plugins hosted via HTTP/S.

**Why is this needed**:

Currently index files can only be retrieved via HTTP/S (no authentication). This is useful for public rule sets, but private rule sets would require both some sort of authentication support and ideally support for different more cloud native storage backends.

max-frank:
https://github.com/falcosecurity/falcoctl/pull/301 would be my suggested first step to implementing/making it easy for the community to add support for new index backend support.

max-frank:
Adding comment here as well for visibility
https://github.com/falcosecurity/falcoctl/pull/301#issuecomment-1579826684
> Thinking out loud here some potential future backends might be using HTTP/S as their protocol language but could require more complex requests (i.e., not only a simple GET request).
So instead of adding this URL scheme based approach adding a backend option to index configurations might be better. This could default to simple HTTP/S for backward compatibility and would probably more flexible.

PR #301, opened 2023-06-06
feat(index/fetch) Add index URL schemes support to fetch

max-frank:
Adding different fetchers per URL scheme allows falcoctl indices to be hosted on many different platforms.

**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

> /area cli

> /area tests

> /area examples

To pave way to support index hosting backends other than plain HTTP/S.

First step to implementing #300


pkg/index/fetch.go
```diff
@@ -17,33 +17,47 @@ package index
...
+	"github.com/falcosecurity/falcoctl/pkg/index/fetch/http"
 	"gopkg.in/yaml.v3"
 )
 
-// Fetch retrieves a remote index using its URL.
-func Fetch(ctx context.Context, url, name string) (*Index, error) {
-	req, err := http.NewRequestWithContext(ctx, "GET", url, http.NoBody)
-	if err != nil {
-		return nil, fmt.Errorf("cannot fetch index: %w", err)
+type FetchFunc func(context.Context, string) ([]byte, error)
```
max-frank, line 27: For now I set the return signature to `[]byte` to keep the unmarshall logic in the main fetch function. Incase it is desired to support backends that store indices not as single YAML file we could also move the unmarshal logic into a util function and have the `Fetch` functions return an index entry list instead.

max-frank:
Thinking out loud here some potential future backends might be using HTTP/S as their protocol language but could require more complex requests (i.e., not only a simple GET request).
So instead of adding this URL scheme based approach adding a `backend` option to index configurations might be better. This could default to simple HTTP/S for backward compatibility and would probably more flexible.

max-frank:
> Thinking out loud here some potential future backends might be using HTTP/S as their protocol language but could require more complex requests (i.e., not only a simple GET request). So instead of adding this URL scheme based approach adding a `backend` option to index configurations might be better. This could default to simple HTTP/S for backward compatibility and would probably more flexible.

changed the approach to this

cmd/add_test.go
```diff
@@ -0,0 +1,122 @@
...
+				"\"notaprotocal://something\": unsupported protocol scheme \"notaprotocal\"")
+		})
+
+		When("with invalid backend", func() {
+			BeforeEach(func() {
+				args = []string{indexCmd, addCmd, "--config", configFile, indexName, "http://noindex", "notabackend"}
+			})
+			addAssertFailedBehavior(indexAddUsage, "ERRO: unable to add index: unable to fetch index \"testName\" "+
+				"with URL \"http://noindex\": unsupported index backend type: notabackend")
+		})
```
max-frank, line 119: I only added failure tests for the new behavior atm, because i noticed that the current implementation of
`falcoctl index add` behaves strangely or could be considered buggy i.e.,

If you call `index add` for and `index name` for which there already exists in the `cache` the CMD will **always** return `Index %q successfully added` and do no modification. For example  consider the following sequence

```
$ cat example.yaml                                                                                      
cat: example.yaml: No such file or directory
$ ./falcoctl index add --config example.yaml example https://falcosecurity.github.io/falcoctl/index.yaml
 INFO  Adding index
 INFO  Index "example" successfully added
$ cat example.yaml                                                                                      
indexes:
    - name: falcosecurity
      url: https://falcosecurity.github.io/falcoctl/index.yaml
      backend: ""
    - name: example
      url: https://falcosecurity.github.io/falcoctl/index.yaml
      backend: ""
$ ./falcoctl index add --config example.yaml example NOPE://NOPE                                        
 INFO  Adding index
 INFO  Index "example" successfully added
$ cat example.yaml                                                                                      
indexes:
    - name: falcosecurity
      url: https://falcosecurity.github.io/falcoctl/index.yaml
      backend: ""
    - name: example
      url: https://falcosecurity.github.io/falcoctl/index.yaml
      backend: ""
 ```

 One would expect the add command to error that the index already exists or that the index is invalid.

**Comment by max-frank** [comment]
I was playing around with the follow up PR that would add GCS support on top of the changes made here and noticed that we kinda need to pass the index entry config to the fetch function. Since if we dont getting the backend config for an entry becomes tedious. So I updated the PR, but since the fetch logic now references the index config entry struct i had to move it to a separate package to avoid import loops.

With that also came some linting required changes e.g., `ConfigEntry` -> `Entry` since golint does not like if after importing stuff is called `config.ConfigEntry`

File: cmd/add_test.go (hunk 2 above)
Comment (alacuku) [reply cmd/add_test.go:119]: Hi @max-frank, the initial idea was to have the `add command `return an error only in cases when there is not possible to add an index. I agree with you that the message should be different if the index already exists, but it should not error. 

Anyway, feel free to modify the current implementation and extend the tests.

Comment (max-frank) [reply cmd/add_test.go:119]: Understood if its modified behavior I can modify the output in a follow up PR.

I'll try to add more tests relevant to this PR I initial did not because of confusion regarding the above interaction and that I a wall in making the cache and config clean up work correctly in between test runs.

**Comment by alacuku** [comment]
Hi @max-frank ,

> So instead of adding this URL scheme based approach adding a `backend` option to index configurations might be better. This could default to simple HTTP/S for backward compatibility and would probably more flexible.

Could you elaborate more on the `backend` approach? Starting from the URL we can infer the `backend` without the need for users to know the `backend` storing the index. If there is any use case for which the `backend` approach makes the implementation easier I would consider, otherwise let's stick with the  URL scheme.

**Comment by max-frank** [comment]
> Could you elaborate more on the `backend` approach? Starting from the URL we can infer the `backend` without the need for users to know the `backend` storing the index. If there is any use case for which the `backend` approach makes the implementation easier I would consider, otherwise let's stick with the URL scheme.

I was mostly thinking of storage backends that use the same protocol (thus same URI scheme), but work differently e.g., 

in case of HTTP we can have a simple server/API that expects a simple GET request with maybe basic auth, oauth, or some authorization token. That minus the Auth would be the currently supported logic.

Now we can take Azure Blob Storage this can be accessed via HTTP as well, but is a lot more complicated. Looking at the [documentation of the REST API](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob?tabs=azure-ad) it requires certain headers to be set and authorized access uses [special Azure specific authorization types](https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-requests-to-azure-storage) some simple like shared keys some less so with dynamic signatures that have to be put in the header.

Stacking support for these kinds of things into a single scheme would get complicated quickly I believe. Though I am aware that I am thinking a bit ahead here in terms of considering support for these things.

File: cmd/add_test.go (hunk 2 above)
Comment (max-frank) [reply cmd/add_test.go:119]: @alacuku For the tests i have run into a slight problem here. For the add tests to be stable and isolated I need to make sure to clean all index config and cache files after each run, but the location of those files is right now not configurable (see [code here](https://github.com/falcosecurity/falcoctl/blob/main/internal/config/config.go#L149-L152)). 

Now since I cant use an isolated tmp config dir it would mean that `index add` tests would have side effects on developer machines, i.e., Indices from tests are added to their real local configuration and if we were to do a clean up after a test case we would delete all their index config.

Is it ok if i skip the positive tests here for now until this issue is addressed?

**Comment by alacuku** [comment]
> Now we can take Azure Blob Storage this can be accessed via HTTP as well, but is a lot more complicated. Looking at the [documentation of the REST API](https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob?tabs=azure-ad) it requires certain headers to be set and authorized access uses [special Azure specific authorization types](https://learn.microsoft.com/en-us/rest/api/storageservices/authorize-requests-to-azure-storage) some simple like shared keys some less so with dynamic signatures that have to be put in the header.

It would be great to have them both. If the user does not explicitly set the backend `falcoctl` will try to guess it based on the URL scheme. For URL schemes shared by different backends, it would ask for the backend to be set explicitly by the user.

Do you think we can do it in this PR?

**Comment by max-frank** [comment]
> It would be great to have them both. If the user does not explicitly set the backend `falcoctl` will try to guess it based on the URL scheme. For URL schemes shared by different backends, it would ask for the backend to be set explicitly by the user.
> 
> Do you think we can do it in this PR?

I can add some simple scheme to backend mapping with one default for each scheme.

**Comment by max-frank** [comment]
> I can add some simple scheme to backend mapping with one default for each scheme.
Added the logic for now its just http and https mapping to the same string, but once there are more backends it will not look as pointless e.g., for Google Cloud Storage `gs` -> `gcs` etc.

File: internal/utils/indexes.go [hunk 3]
```diff
@@ -19,10 +19,11 @@ import (
 	"path/filepath"
 
 	"github.com/falcosecurity/falcoctl/pkg/index"
+	"github.com/falcosecurity/falcoctl/pkg/index/config"
 )
 
 // Indexes returns the merge of all configured indexes.
-func Indexes(indexConfig *index.Config, path string) (*index.MergedIndexes, error) {
+func Indexes(indexConfig *config.Config, path string) (*index.MergedIndexes, error) {
```
alacuku, line 26: Seems to be dead code, It would be nice to remove it. I think it's a leftover from previous refactors.

alacuku (commented):
Left two minor comments. It seems ok to me. We are a rebase away from merging this one. 

Nice job @max-frank! :heart:

max-frank:
Comment updated, dead code removed and rebased to latest main 👍

max-frank:
Also I've been working offline on the followup PR to add GCS as backend. That one will be ready for review soon as well.

alacuku (approved):
/LGTM

alacuku:
> Also I've been working offline on the followup PR to add GCS as backend. That one will be ready for review soon as well.

I would be happy to review it. And again, thank you for your efforts in the falcoctl project.

FedeDP (approved):
LGTM, thanks!
Thanks also @alacuku for the review!
/approve

PR #309, opened 2023-07-27
new: introduce `asset` artifact type

loresuso:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

/area library

/area cli

> /area tests

> /area examples

This PR aims at introducing a new artifact type that `falcoctl` can handle named `asset`. 
This kind of artifact is something that can either be consumed by Falco plugins or Falco itself but it's neither a plugin nor a rulesfile.

Fixes #

### DIFF
--- a/cmd/registry/push/push_test.go
+++ b/cmd/registry/push/push_test.go
@@ -45,7 +45,7 @@ Flags:
       --platform stringArray       os and architecture of the artifact in OS/ARCH format (only for plugins artifacts)
   -r, --requires stringArray       set an artifact requirement (can be specified multiple times). Example: "--requires plugin_api_version:1.2.3"
   -t, --tag stringArray            additional artifact tag. Can be repeated multiple times
-      --type ArtifactType          type of artifact to be pushed. Allowed values: "rulesfile", "plugin" (default )
+      --type ArtifactType          type of artifact to be pushed. Allowed values: "rulesfile", "plugin", "asset" (default )
       --version string             set the version of the artifact
 
 Global Flags:
@@ -99,7 +99,7 @@ Flags:
       --platform stringArray       os and architecture of the artifact in OS/ARCH format (only for plugins artifacts)
   -r, --requires stringArray       set an artifact requirement (can be specified multiple times). Example: "--requires plugin_api_version:1.2.3"
   -t, --tag stringArray            additional artifact tag. Can be repeated multiple times
-      --type ArtifactType          type of artifact to be pushed. Allowed values: "rulesfile", "plugin"
+      --type ArtifactType          type of artifact to be pushed. Allowed values: "rulesfile", "plugin", "asset"
       --version string             set the version of the artifact
 
 Global Flags:
@@ -192,10 +192,10 @@ var registryPushTests = Describe("push", func() {
 
 		When("multiple rulesfiles", func() {
 			BeforeEach(func() {
-				args = []string{registryCmd, pushCmd, rulesRepo, "--config", configFile, rulesfiletgz, rulesfiletgz,
-					"--type", "rulesfile", "--version", "1.1.1", "--plain-http"}
+				args = []string{registryCmd, pushCmd, "--config", configFile,
+					"--type", "rulesfile", "--version", "1.1.1", "--plain-http", rulesRepo, rulesfiletgz, rulesfiletgz}
 			})
-			pushAssertFailedBehavior(registryPushUsage, "ERROR expecting 1 rulesfile object received 2: invalid number of rulesfiles")
+			pushAssertFailedBehavior(registryPushUsage, "ERROR expecting 1 rulesfile object, received 2: invalid number of rulesfiles")
 		})
 
 		When("unreachable registry", func() {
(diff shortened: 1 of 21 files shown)


pkg/oci/pusher/pusher.go
```diff
@@ -163,8 +167,8 @@ func (p *Pusher) Push(ctx context.Context, artifactType oci.ArtifactType,
 		}
 	}
 
-	if artifactType == oci.Rulesfile {
-		// We should have only one manifestDesc.
+	if artifactType != oci.Plugin {
```
FedeDP, line 170: I'd make the check more explicit here (ie:
```suggestion
	if artifactType == oci.Rulesfile || artifactType == oci.Asset {
```

loresuso: True, it's better being explicit if we add another one too!

loresuso:
cc @alacuku

pkg/options/directory_options.go
alacuku, line 1: Since the `directory_options.go` file is under the `options` package we could drop the `_options` suffix. The package name is self-explanatory.

internal/utils/compress.go
```diff
@@ -75,20 +71,61 @@ func CreateTarGzArchive(path string) (file string, err error) {
 		}
 	}()
 
+	if fInfo.IsDir() {
```
alacuku, line 74: Would be nice to have some unit tests.

internal/utils/extract.go
```diff
@@ -48,14 +51,28 @@ func ExtractTarGz(gzipStream io.Reader, destDir string) ([]string, error) {
...
 		case tar.TypeDir:
-			return nil, fmt.Errorf("unexepected dir inside the archive, expected to find only files without any tree structure")
-		case tar.TypeReg:
-			if strings.Contains(header.Name, "..") {
-				return nil, fmt.Errorf("not allowed relative path in tar archive")
+			if artifactType == oci.Plugin || artifactType == oci.Rulesfile {
+				return nil, fmt.Errorf("unexepected dir inside the archive, "+
+					"expected to find only files without any tree structure for %q artifacts", artifactType.String())
+			} else if artifactType == oci.Asset {
+				if dirs > 1 {
```
alacuku, line 64: Why do we need this check? Some comments and unit tests for this case would be helpful.

alacuku (commented):
Hey @loresuso, thanks for the PR.

Left some comments.

pkg/options/directory_options.go
loresuso: This is true, but we have the same issue with all the files in this directory. Also, I'd drop the `options` also from the struct name, I will open a refactoring PR soon

internal/utils/compress.go
loresuso: Ok I can do it!

loresuso:
I think this is ready for another round of review. It should be all covered by tests for pushing assets in `pusher_test.go` and the out compressing function `createTargZ` in `compress_test.go`. Do we need more?

cmd/registry/push/push_test.go
```diff
@@ -99,7 +99,7 @@ Flags:
       --platform stringArray       os and architecture of the artifact in OS/ARCH format (only for plugins artifacts)
   -r, --requires stringArray       set an artifact requirement (can be specified multiple times). Example: "--requires plugin_api_version:1.2.3"
   -t, --tag stringArray            additional artifact tag. Can be repeated multiple times
-      --type ArtifactType          type of artifact to be pushed. Allowed values: "rulesfile", "plugin"
+      --type ArtifactType          type of artifact to be pushed. Allowed values: "rulesfile", "plugin", "asset"
```
alacuku, line 102: It would be nice to implement the flag using the enum here: https://github.com/falcosecurity/falcoctl/blob/main/pkg/options/enum.go

alacuku (commented):
Hey @loresuso, it seems ok to me. Left a suggestion.

cmd/registry/push/push_test.go
loresuso: oci.ArtifactType already implements the cobra `Value` interface. I prefer this way instead of a custom one, if I understood correctly

alacuku: The `enum` type would be more suitable for such a use case. Up to you if you want to implement it or not.

loresuso: I will introduce this in a second PR to speed up this one entering the codebase

leogr:
@loresuso can you rebase this PR, please?

loresuso:
@leogr the conflict involves also the new driver logic, I'm trying to figure how to deal with that

pkg/driver/distro/distro.go
```diff
@@ -319,6 +319,9 @@ func downloadKernelSrc(ctx context.Context,
 		return env, err
 	}
 
+	// todo(loresuso,fededp): use oci.Asset as the artifact type
```
FedeDP, line 322: Cleanup this comment?

internal/utils/extract.go
```diff
@@ -98,3 +98,29 @@ func stripComponents(headerName string, stripComponents int) string {
 	}
 	return filepath.Clean(strings.Join(names[stripComponents:], "/"))
 }
+
+func listHeaders(gzipStream io.Reader) ([]string, error) {
```
FedeDP, line 102: Is this utility used somewhere else aside tests?
Otherwise i'd move it inside internal/utils/compress_test.go

loresuso: agreed, done!

FedeDP (approved):
/approve

Issue #327, opened 2023-09-27
Implement driver loading capabilities in falcoctl


LucaGuerra:
**What would you like to be added**:

According to this proposal: https://github.com/falcosecurity/falco/blob/master/proposals/20221129-artifacts-distribution.md#deprecate-falco-driver-loader we need a new home for the Falco driver loader. This issue acts as a basic requirement specification for this and as a place for discussion. The issue will be updated with more details as it's being developed.

This issue covers feature parity with the driver loader script. The main required features are:
* Distribution detection
* Ability to download prebuilts from https://download.falco.org
* Ability to compile with the local toolchain (i.e. the one available in the container image or the one in the host)
* Options to select compilation, download or both with fallback
* Options to select kmod or (classic) bpf probe -- the modern ebpf probe does not need a driver loader
* Support for at least debian, ubuntu, amzn, flatcar, minikube, bottlerocket, talos and generic

In addition, the design should allow for other projects using the Falco libs to use falcoctl or import it into their own Go-based tooling, meaning that their strategy to download and build prebuilt could be a little different and the design should still allow for that to work. In addition, it should be easier for contributors to add support for their own favorite distribution, since it has proven to be a bit hard to do the same in the bash script.

**Why is this needed**:

https://github.com/falcosecurity/falco/blob/master/proposals/20221129-artifacts-distribution.md#deprecate-falco-driver-loader

FedeDP:
Still missing:
* tests
* fix up driver type names to match https://github.com/falcosecurity/falco/pull/2413 once it get merged (eg: `modern-bpf` will become `modern-ebpf`) and eventually other Falco config changes
* Improvements, eg: use `unix.lsmod/insmod` etc etc instead of `exec.Command` to reduce exec.Command usages

FedeDP:
I'd like to tag a 0.7.0-alpha once https://github.com/falcosecurity/falco/pull/2413 is merged and needed changes land in falcoctl, to be able to test Falco with the new driver-loader.

leogr:
> I'd like to tag a 0.7.0-alpha once [falcosecurity/falco#2413](https://github.com/falcosecurity/falco/pull/2413) is merged and needed changes land in falcoctl, to be able to test Falco with the new driver-loader.

I'll take a look at https://github.com/falcosecurity/falco/pull/2413 soon :+1:

FedeDP:
IMHO we can close this one; Falco master now ships the new driver loader, and tests are ongoing.
[...]

PR #343, opened 2023-11-03
new: driver command

FedeDP:
**What type of PR is this?**

/kind feature

/area library
/area cli

First draft at implementing `falcoctl driver` command, see #327 

Eventually #327 will get fixed, but i don't think a single PR will make it :)

Fixes #

What we need:
- [x] `select` subcmd with possibility for an "autoselection" smart logic in place; this is not required by Falco but might be useful for other consumers of the falcoctl libraries and for us in the future
- [x] `select` stores the currently selected driver in the falcoctl config
- [x] `select` must be able to communicate to Falco the driver it must use (we need a patch to allow Falco config to specify a driver to be used) -> https://github.com/falcosecurity/falco/pull/2413 -> this is implemented in falcoctl, even if will need some more fixes
- [ ] `select` tests
- [x] `prepare` skeleton (this will be the step where we either download or build the drivers)
- [x] ~~autoload of target `driverversion` from running Falco process, if any~~ Retired with https://github.com/falcosecurity/falcoctl/pull/343#issuecomment-1803345048
- [x] `prepare` download  
- [x] `prepare` build
- [x] specific distros support (similar to the [`get_target_id`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L104) falco-driver-loader function)
- [ ] `prepare` tests
- [x] Support options that were supported by falco-driver-loader:
- - [x] `echo "  DRIVER_INSECURE_DOWNLOAD whether you want to allow insecure downloads or not"`
- - [x] ~~`echo "  DRIVER_CURL_OPTIONS      specify additional options to be passed to curl command used to download Falco drivers"`~~ -> **DEPRECATED**
- - [x] `echo "  DRIVER_KERNEL_RELEASE    specify the kernel release for which to download/build the driver in the same format used by 'uname -r' (e.g. '6.1.0-10-cloud-amd64')"`
- - [x] `echo "  DRIVER_KERNEL_VERSION    specify the kernel version for which to download/build the driver in the same format used by 'uname -v' (e.g. '#1 SMP PREEMPT_DYNAMIC Debian 6.1.38-2 (2023-07-27)')"`

### DIFF
--- a/cmd/driver/config/config.go
+++ b/cmd/driver/config/config.go
@@ -1,0 +1,246 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 The Falco Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package driverconfig
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/pterm/pterm"
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/falcosecurity/falcoctl/internal/config"
+	"github.com/falcosecurity/falcoctl/internal/utils"
+	driverdistro "github.com/falcosecurity/falcoctl/pkg/driver/distro"
+	driverkernel "github.com/falcosecurity/falcoctl/pkg/driver/kernel"
+	drivertype "github.com/falcosecurity/falcoctl/pkg/driver/type"
+	"github.com/falcosecurity/falcoctl/pkg/options"
+)
+
+const (
+	configMapDriverTypeKey = "driver_mode"
+)
+
+type driverConfigOptions struct {
+	*options.Common
+	Type       *options.DriverTypes
+	Version    string
+	Repos      []string
+	Name       string
+	HostRoot   string
+	Update     bool
+	Namespace  string
+	KubeConfig string
+}
+
+// NewDriverConfigCmd configures a driver and stores it in config.
+func NewDriverConfigCmd(ctx context.Context, opt *options.Common) *cobra.Command {
+	o := driverConfigOptions{
+		Common: opt,
+		Type:   options.NewDriverTypes(),
+	}
+
+	cmd := &cobra.Command{
+		Use:                   "config [flags]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Configure a driver",
+		Long:                  "Configure a driver for future usages with other driver subcommands",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.RunDriverConfig(ctx, cmd)
+		},
+	}
+
+	cmd.Flags().Var(o.Type, "type", "Driver type to be configured "+o.Type.Allowed())
+	cmd.Flags().StringVar(&o.Version, "version", config.DefaultDriver.Version, "Driver version to be configured.")
+	cmd.Flags().StringSliceVar(&o.Repos, "repo", config.DefaultDriver.Repos, "Driver repo to be configured.")
+	cmd.Flags().StringVar(&o.Name, "name", config.DefaultDriver.Name, "Driver name to be configured.")
+	cmd.Flags().StringVar(&o.HostRoot, "host-root", config.DefaultDriver.HostRoot, "Driver host root to be configured.")
+	cmd.Flags().BoolVar(&o.Update, "update-falco", true, "Whether to update Falco config/configmap.")
+	cmd.Flags().StringVar(&o.Namespace, "namespace", "", "Kubernetes namespace.")
+	cmd.Flags().StringVar(&o.KubeConfig, "kubeconfig", "", "Kubernetes config.")
+	return cmd
+}
+
+// RunDriverConfig implements the driver configuration command.
+func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Command) error {
+	var (
+		dType drivertype.DriverType
+		err   error
+	)
+
+	driverCfg, err := config.Driverer()
+	if err != nil {
+		return err
+	}
+
+	loggerArgs := make([]pterm.LoggerArgument, 0)
+	if f := cmd.Flags().Lookup("version"); f != nil && f.Changed {
+		driverCfg.Version = o.Version
+		loggerArgs = append(loggerArgs, pterm.LoggerArgument{
+			Key:   "driver version",
+			Value: o.Version,
+		})
+	}
+	if f := cmd.Flags().Lookup("repo"); f != nil && f.Changed {
+		driverCfg.Repos = o.Repos
+		loggerArgs = append(loggerArgs, pterm.LoggerArgument{
+			Key:   "driver repos",
+			Value: strings.Join(o.Repos, ","),
+		})
+	}
+	if f := cmd.Flags().Lookup("name"); f != nil && f.Changed {
+		driverCfg.Name = o.Name
+		loggerArgs = append(loggerArgs, pterm.LoggerArgument{
+			Key:   "driver name",
+			Value: o.Name,
+		})
+	}
+	if f := cmd.Flags().Lookup("host-root"); f != nil && f.Changed {
+		driverCfg.HostRoot = o.HostRoot
+		loggerArgs = append(loggerArgs, pterm.LoggerArgument{
+			Key:   "driver host root",
+			Value: o.HostRoot,
+		})
+	}
+	if f := cmd.Flags().Lookup("type"); f != nil && f.Changed {
+		loggerArgs = append(loggerArgs, pterm.LoggerArgument{
+			Key:   "driver type",
+			Value: o.Type.String(),
+		})
+		if o.Type.String() != "auto" {
+			// Ok driver type was enforced by the user
+			dType, err = drivertype.Parse(o.Type.String())
+			if err != nil {
+				return err
+			}
+		} else {
+			// automatic logic
+			info, err := driverkernel.FetchInfo("", "")
+			if err != nil {
+				return err
+			}
+			o.Printer.Logger.Debug("Fetched kernel info", o.Printer.Logger.Args(
+				"arch", info.Architecture.ToNonDeb(),
+				"kernel release", info.String(),
+				"kernel version", info.KernelVersion))
+
+			d, err := driverdistro.DiscoverDistro(info, driverCfg.HostRoot)
+			if err != nil {
+				return err
+			}
+			o.Printer.Logger.Debug("Discovered distro", o.Printer.Logger.Args("target", d))
+
+			dType = d.PreferredDriver(info)
+			if dType == nil {
+				return fmt.Errorf("automatic driver selection failed")
+			}
+		}
+		driverCfg.Type = dType
+	}
+
+	o.Printer.Logger.Info("Running falcoctl driver config", loggerArgs)
+
+	if o.Update {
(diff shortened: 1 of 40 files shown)


FedeDP:
About `select` smart autologic; this is as simple as adding new "distro" overriding the `PreferredDriver` interface method, like this:
```
func init() {
	distros["arch"] = &arch{generic: &generic{}}
}

type arch struct {
	*generic
}

func (c *arch) PreferredDriver(info kernel.Info) _type.DriverType {
        if info.Architecture == "aarch64" {
            dtype, _ := _type.Parse("bpf")
            return dtype
        }
        dtype, _ := _type.Parse("modern-bpf")
        return dtype
}
```
(ok this is a super stupid example btw!)

FedeDP:
Example output:
```
sudo ./falcoctl driver prepare --driver-version 6.0.1+driver
[sudo] password di federico:
2023-11-03 15:31:36 INFO  Running falcoctl driver prepare
                      ├ driver version: 6.0.1+driver
                      ├ driver type: kmod
                      ├ arch: x86_64
                      ├ kernel release: 6.5.7-arch1-1
                      └ kernel version: 1
2023-11-03 15:31:36 INFO  found distro target: arch
2023-11-03 15:31:36 INFO  Trying to download a driver url: https://download.falco.org/driver/6.0.1+driver/x86_64/falco_arch_6.5.7-arch1-1_1.ko
2023-11-03 15:31:36 ERROR unable to find a prebuilt falco driver
2023-11-03 15:31:36 ERROR build unimplemented
```

Makefile
```diff
@@ -56,7 +56,7 @@ fmt: gci addlicense
 	go mod tidy
 	go fmt ./...
 	find . -type f -name '*.go' -a -exec $(GCI) write -s standard -s default -s "prefix(github.com/falcosecurity/falcoctl)" {} \;
-	find . -type f -name '*.go' -exec $(ADDLICENSE) -l apache -c "The Falco Authors" -y "$(shell date +%Y)" {} \;
+	find . -type f -name '*.go' -exec $(ADDLICENSE) -l apache -s -c "The Falco Authors" -y "$(shell date +%Y)" {} \;
```
FedeDP, line 59:
```
  -s    Include SPDX identifier in license header. Set -s=only to only include SPDX identifier.
```

cmd/driver/driver_linux.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/spf13/cobra"
+
+	driverprepare "github.com/falcosecurity/falcoctl/cmd/driver/prepare"
+	driverselect "github.com/falcosecurity/falcoctl/cmd/driver/select"
+	"github.com/falcosecurity/falcoctl/internal/config"
+	commonoptions "github.com/falcosecurity/falcoctl/pkg/options"
+)
+
+// NewDriverCmd returns the driver command.
+func NewDriverCmd(ctx context.Context, opt *commonoptions.Common) *cobra.Command {
```
FedeDP, line 32: `driver` subcmd is only available on linux systems.

cmd/driver/driver_others.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/spf13/cobra"
+	"golang.org/x/net/context"
+
+	commonoptions "github.com/falcosecurity/falcoctl/pkg/options"
+)
+
+// NewDriverCmd returns an empty driver command since it is not supported on non linuxes
+func NewDriverCmd(ctx context.Context, opt *commonoptions.Common) *cobra.Command {
+	return &cobra.Command{}
```
FedeDP, line 29: By returning an empty command, cobra will skip it.

cmd/driver/prepare/prepare.go
```diff
@@ -0,0 +1,131 @@
...
+		"kernel version", info.KernelVersion))
+
+	if !driver.Type.HasArtifacts() {
+		o.Printer.Logger.Info("no artifacts needed")
+		return nil
+	}
+
+	d, err := driverdistro.DiscoverDistro(o.Printer, o.HostRoot)
+	if err != nil {
+		if errors.Is(err, driverdistro.ErrUnsupported) && o.Build {
```
FedeDP, line 100: `errUnsupported` returns a generic distro, to attempt a build.

FedeDP:
What is implemented so far:
* [get_target_id](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L104) magic
* kmod/bpf download
* [clean_kernel_module](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L374) magic
* [mounting of debugfs for bpf](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L630)

What we miss:
* build kmod
* build bpf
* `cos_version_greater`
* `get_kernel_config`
* `flatcar_relocate_tools`
* `insmod`/`modprobe` of kmod once built/downloaded
* `DRIVER_INSECURE_DOWNLOAD`, `DRIVER_CURL_OPTIONS`, `DRIVER_KERNEL_RELEASE`, `DRIVER_KERNEL_VERSION` override


Open points:
* who will pass `--driver-version` parameter? Falco knows it, but falcoctl does not (unless we let it call the Falco endpoint to gather the info, but since `falcoctl driver` needs to run **before** Falco... 
* how to support [`source_only`])https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L685) and [`print_env`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L686) falco-driver-loader options?

FedeDP:
I think this is already big enough as a starting point.
/cc @alacuku @leogr

FedeDP:
Regarding deprecated options, we will drop:
* [`--source-only`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L685) support
* [`BPF_USE_LOCAL_KERNEL_SOURCES`](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L559) suppport

FedeDP:
Regarding
> autoload of target `driverversion` from running Falco process, if any

i added a small trick to autoload a driverversion if left empty, that could be useful: basically, we expect falcosecurity driver sources to be available under `/usr/src/falco-$driverversion`. 
What i did is to cycle on any available `/usr/src/falco-*+driver` folder, and load the greatest semver present.
Normally, falco-driver-loader image is shipped with just a single driver, but in any case we will use the most recent.

With this small trick we can ensure that `falcoctl driver` is able to do its magic without any driverversion specified.

FedeDP:
`printenv` examples vs `falco-driver-loader` output:
```
sudo ./falcoctl driver printenv
DRIVER="kmod"
DRIVERS_REPO="https://download.falco.org/driver"
DRIVER_VERSION="6.0.1+driver"
DRIVER_NAME="falco"
TARGET_ID="ubuntu-aws"
ARCH="x86_64"
KERNEL_RELEASE="6.2.0-1012-aws"
KERNEL_VERSION="12~22.04.1"
```
vs
```
./falco-driver-loader --print-env
ARCH="x86_64"
KERNEL_RELEASE="6.2.0-1012-aws"
KERNEL_VERSION="12~22.04.1"
ENABLE_COMPILE="yes"
ENABLE_DOWNLOAD="yes"
TARGET_ID="ubuntu-aws"
DRIVER="module"
DRIVERS_REPO="https://download.falco.org/driver"
DRIVER_VERSION="6.0.1+driver"
DRIVER_NAME="falco"
FALCO_VERSION="0.37.0-86+7319b93"
```

The most important thing to check here is that `KERNEL_RELEASE`, `KERNEL_VERSION` and `TARGET_ID` match, because it means that our per-distro logic is working fine.
We lose: `ENABLE_COMPILE`, `ENABLE_DOWNLOAD` and `FALCO_VERSION` (the formers make no sense, the latter cannot be retrieved). `DRIVER`  becomes `kmod` instead of `module`. This can be easily reverted but i feel `kmod` is a better name, and we are using it elsewhere.

cmd/driver/config/config.go
```diff
@@ -0,0 +1,219 @@
...
+	err := replaceDriverTypeInFalcoConfig(hostRoot+"/etc/falco/falco.yaml", driverType)
+	if err == nil {
+		return nil
+	}
+	if !os.IsNotExist(err) {
+		return err
+	}
+	// falco config does not exist; try K8S.
+
+	// TODO double check
```
FedeDP, line 178: This is a big TODO

FedeDP:
Al2023 example:
```
sudo ./falcoctl driver printenv
DRIVER="kmod"
DRIVERS_REPO="https://download.falco.org/driver"
DRIVER_VERSION=""
DRIVER_NAME="falco"
HOST_ROOT=""
TARGET_ID="amazonlinux2023"
ARCH="x86_64"
KERNEL_RELEASE="6.1.59-84.139.amzn2023.x86_64"
KERNEL_VERSION="1"
```
vs:
```
sudo ./falco-driver-loader --print-env
ARCH="x86_64"
KERNEL_RELEASE="6.1.59-84.139.amzn2023.x86_64"
KERNEL_VERSION="1"
ENABLE_COMPILE="yes"
ENABLE_DOWNLOAD="yes"
TARGET_ID="amazonlinux2023"
DRIVER="module"
DRIVERS_REPO="https://download.falco.org/driver"
DRIVER_VERSION="6.0.1+driver"
DRIVER_NAME="falco"
FALCO_VERSION="0.37.0-87+8cfa79f"
```
```

**Review by alacuku — COMMENTED** [review]
Thanks for the PR, left some comments.

File: cmd/driver/config/config.go [hunk 6]
```diff
@@ -0,0 +1,219 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.RunDriverConfig(cmd)
+		},
+	}
+
+	cmd.Flags().StringVar(&o.Type, "type", "", "Driver type to be configured. Allowed: "+strings.Join(drivertype.GetDriverTypes(), ",")+",auto")
+	cmd.Flags().StringVar(&o.Version, "version", "", "Driver version to be configured.")
+	cmd.Flags().StringSliceVar(&o.Repos, "repo", nil, "Driver repo to be configured.")
+	cmd.Flags().StringVar(&o.Name, "name", "", "Driver name to be configured.")
+	cmd.Flags().StringVar(&o.HostRoot, "host-root", "", "Driver host root to be configured.")
```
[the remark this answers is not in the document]
leogr: We should be able to set defaults for `type`, `repo`, and `name`. 
We can't set a default for `host-root` (its value usually changes depending on the context). Also, we should not set a default for `version`, since it may change at any time. 

Does it make sense?

internal/config/config.go
```diff
@@ -158,6 +174,24 @@ type Install struct {
...
+
+// Driver represents the resolved driver configuration, exposed to be consumed.
+type Driver struct {
+	Type     drivertype.DriverType
+	Name     string
+	Repos    []string
+	Version  string
+	HostRoot string
+}
+
```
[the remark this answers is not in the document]
FedeDP: It is, i mean it is the `driver` config with Type resolved to its interface type.
I can also add a `pkg/driver/types` package but it seems overkill IMHO.

cmd/driver/config/config.go
```diff
@@ -0,0 +1,219 @@
...
+		Use:                   "config [flags]",
+		DisableFlagsInUseLine: true,
+		Short:                 "Configure a driver",
+		Long:                  "Configure a driver for future usages with other driver subcommands",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.RunDriverConfig(cmd)
+		},
+	}
+
+	cmd.Flags().StringVar(&o.Type, "type", "", "Driver type to be configured. Allowed: "+strings.Join(drivertype.GetDriverTypes(), ",")+",auto")
```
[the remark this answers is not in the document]
FedeDP: Mmh i don't think it is fitting in this case (i mean, just to get a list of driver types strings).

FedeDP:
I consider this PR ready!
I will add tests and eventually improve things (together with @alacuku perhaps!) in next PRs (yes there will surely be other PRs) :smile:

FedeDP:
Squashed to a single commit :) @alacuku is now happy :smiley:

alacuku (approved):
/LGTM

pkg/driver/distro/distro.go
```diff
@@ -0,0 +1,340 @@
...
+const (
+	// DefaultFalcoRepo is the default repository provided by falcosecurity to download driver artifacts from.
+	kernelDirEnv            = "KERNELDIR"
+	kernelSrcDownloadFolder = "kernel-sources"
+)
+
+var distros = map[string]Distro{}
+
+// ErrUnsupported is the error returned when the target distro is not supported.
+var ErrUnsupported = fmt.Errorf("failed to determine distro")
```
maxgio92, line 48: `errors.New()` instead? 🤓

FedeDP: Uh yep you are right :D

FedeDP: Solved in latest force-push!

PR #353, opened 2023-11-22
cleanup(pkg/driver): use unix where available.

FedeDP:
**What type of PR is this?**

/kind cleanup

/area library

Small cleanups on top of #343 : use `unix` golang library where available.
NOTE: reimplementing `insmod` or `modprobe` in go, instead of executing it through an `exec.Command` is pretty convoluted since they do quite lot of things.
See for example how is that implemented in this library: https://github.com/pmorjan/kmod/tree/master
Also, that library itself does not help us much since the job is still pretty complex.
I'd avoid the risk of introducing bugs reimplementing ourselves 2 of the most widely used bash utilities.

Fixes #

### DIFF
--- a/pkg/driver/type/kmod.go
+++ b/pkg/driver/type/kmod.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/falcosecurity/driverkit/pkg/kernelrelease"
 	"golang.org/x/net/context"
+	"golang.org/x/sys/unix"
 
 	"github.com/falcosecurity/falcoctl/pkg/output"
 )
@@ -37,14 +38,6 @@ const (
 	maxRmmodWait  = 10
 	rmmodWaitTime = 5 * time.Second
 )
-
-type errMissingDep struct {
-	program string
-}
-
-func (e *errMissingDep) Error() string {
-	return fmt.Sprintf("This program requires %s.", e.program)
-}
 
 func init() {
 	driverTypes[TypeKmod] = &kmod{}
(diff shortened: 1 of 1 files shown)


pkg/driver/type/kmod.go
```diff
@@ -61,40 +54,45 @@ func (k *kmod) String() string {
...
-	_, err = exec.Command("bash", "-c", "hash rmmod").Output()
-	if err != nil {
-		return &errMissingDep{program: "rmmod"}
-	}
-
 	kmodName := strings.ReplaceAll(driverName, "-", "_")
 	printer.Logger.Info("Check if kernel module is still loaded.")
-	lsmodCmdArgs := fmt.Sprintf(`lsmod | cut -d' ' -f1 | grep -qx %q`, kmodName)
-	_, err = exec.Command("bash", "-c", lsmodCmdArgs).Output() //nolint:gosec // false positive
+	f, err := os.Open("/proc/modules")
```
FedeDP, line 59: Instead of using `lsmod`, manually scan `/proc/modules`.

pkg/driver/type/kmod.go
```diff
@@ -61,40 +54,45 @@ func (k *kmod) String() string {
...
+			line := scanner.Text()
+			fields := strings.Fields(line)
+			if len(fields) > 0 && fields[0] == kmodName {
+				found = true
+				unloaded := false
+				// Module is still loaded, try to remove it
+				for i := 0; i < maxRmmodWait; i++ {
+					printer.Logger.Info("Kernel module is still loaded.")
+					printer.Logger.Info("Trying to unload it with 'rmmod'.")
+					if err = unix.DeleteModule(kmodName, 0); err == nil {
```
FedeDP, line 75: Instead of calling `rmmod` directly, just use `unit.DeleteModule`.

PR #356, opened 2023-11-24
new(pkg): use driverkit `local` build processor instead of implementing drivers build

FedeDP:
**What type of PR is this?**

/kind feature

/area library

Driverkit v0.16.0 added a new `local` build processor: https://github.com/falcosecurity/driverkit/pull/306.
Making use of it, we can drop all the custom logic for drivers build.

Refs https://github.com/falcosecurity/falcoctl/issues/327

Fixes #

TODO:
- [x] bump to driverkit v0.19.1 to fix build
- [x] ~~when dkms build fails, we need some way to get the error, like accessing /var/lib/dkms... log. This is what we currently do in falcoctl main and we should retain the behavior~~ -> this is now done by driverkit itself: https://github.com/falcosecurity/driverkit/pull/342/files
- [x] ~~when build fails, we should print the output at warn/error level like falcoctl does today (eg: https://github.com/falcosecurity/falcoctl/blob/main/pkg/driver/type/bpf.go#L101)~~ -> this is now done by driverkit itself: https://github.com/falcosecurity/driverkit/pull/342/files
- [x] ~~capture `slog` log output (used by driverkit) within our own log level and writer~~ -> no more needed since driverkit itself now uses falcoctl `pkg/output` Printer as logger: https://github.com/falcosecurity/driverkit/pull/333

Also, most of the lines "added" are from go.sum.


FedeDP:
Note: pulling in latest driverkit breaks build because of the usage of:
```
Error: ../../../go/pkg/mod/github.com/falcosecurity/driverkit@v0.16.0/cmd/local.go:90:13: undefined: unix.Utsname
Error: ../../../go/pkg/mod/github.com/falcosecurity/driverkit@v0.16.0/cmd/local.go:91:18: undefined: unix.Uname
```
in it for non-linux builds.
I am not sure whether to drop the dep in driverkit (enforcing user to pass kernelrelease/kernelversion even for `local` builds (it might make sense, after all the 2 parameteres are already mandatory for all others command) or to try something else (don't know what though!)

FedeDP:
/reopen
Ops :D

FedeDP:
Rebased on top of master; moreover:
* it uses driverkit from https://github.com/falcosecurity/driverkit/pull/319
* ran `go get -u` to bump all deps

Only question remaining: driverkit uses `log/slog` package for structured logging (avail since golang 1.21) and internally logs multiple things.
This breaks our `spinner` behavior. Should we:
* either disable driverkit logging (eg: by calling:
```
h := slog.NewTextHandler(io.Discard, &slog.HandlerOptions{})
slog.SetDefault(slog.New(h))
```
ie: by setting an io discarder slog as default slog instance that gets inherited by driverkit.
* or by passing our `printer.Logger.Writer` and `printer.Logger.Level` to the slog instance, like:
```
h := slog.NewTextHandler(printer.Logger.Writer, &slog.HandlerOptions{
		Level: printer.Logger.Level,
		ReplaceAttr: func(groups []string, a slog.Attr) slog.Attr {
			if a.Key == slog.TimeKey {
				return slog.Attr{}
			}
			return a
		}})
slog.SetDefault(slog.New(h))
```
this way, the `slog` used by driverkit will print on our provided writer (without `datetime` attribute because we don't want that).

FedeDP:
Another TODO:
* when build fails, we need some way to get the error, like accessing `/var/lib/dkms...` log. This is what we currently do in falcoctl main.

FedeDP:
I will bump this once:
* https://github.com/falcosecurity/falcoctl/pull/484
* https://github.com/falcosecurity/falcoctl/pull/476
are merged; also i need to make a new release for driverkit once https://github.com/falcosecurity/driverkit/pull/324 is merged.

pkg/driver/distro/distro.go
```diff
@@ -189,48 +192,58 @@ func Build(ctx context.Context,
...
 	if err != nil {
 		return "", err
 	}
 
-	if env == nil {
-		env = make(map[string]string)
-	}
+	srcPath := fmt.Sprintf("/usr/src/%s-%s", driverName, driverVer)
+
+	// TODO shall we call fixupkernel here?
```
FedeDP, line 209: Still unsure: we should build against non fixed up kernel, but eg: debian distro has some very specific fixupkernel logic and i must understand whether we can fix the logic to retrieve kernel urls in driverkit.

pkg/driver/distro/distro.go
```diff
@@ -189,48 +192,58 @@ func Build(ctx context.Context,
...
+	ro.KernelVersion = fixedKr.KernelVersion
+	ro.ModuleDriverName = driverName
+	ro.ModuleDeviceName = driverName
+	ro.KernelRelease = fixedKr.String()
+	ro.Target = d.String()
+	ro.Output = driverType.ToOutput(destPath)
+
+	// Setup slog logger (used by driverkit)
+	// to log on our logger level and writer.
+	programLevel := &slog.LevelVar{}
```
FedeDP, line 223: This will probably be moved to its own helper function.
Basically we want to enforce `slog` logger (used by driverkit) to log at our own log level on our own log writer.

FedeDP:
I need to bump `docker` dep on driverkit to fix the build. So, we will need a 0.18.1 release of driverkit too!

FedeDP:
Driverkit 0.18.1 released: https://github.com/falcosecurity/driverkit/releases/tag/v0.18.1

pkg/driver/distro/distro.go
FedeDP: Double checked: only Ubuntu driverkit builder actually uses KernelVersion; therefore we will probably need to fix it for ubuntu, but other distros are fine using an unfixed kr.

FedeDP:
Bumped to driverkit v0.19.0.
There is a build issue since `pkg/driver/distro` imports `driverkit/cmd` that imports `pkg/options` : https://github.com/falcosecurity/driverkit/blob/master/cmd/config_options.go#L20 and that is an import cycle.
Gonna work on this asap.

FedeDP:
https://github.com/falcosecurity/falcoctl/pull/550 is the first step to fix the build; we will then need to port the changes to https://github.com/falcosecurity/driverkit/pull/342 and tag driverkit v0.19.1. We should be finally fine then.

FedeDP:
Ok driverkit v0.19.1 was tagged; the CI should now be green here, and we can proceed.
@alacuku it would be great if we could merge this before Monday since otherwise i will have to fix all `go.{mod,sum}` conflicts :worried:

alacuku (commented):
/lgtm

I left a suggestion.

pkg/driver/distro/distro.go
```diff
@@ -201,53 +170,67 @@ func Build(ctx context.Context,
...
+	printer.Logger.Info("Trying to compile the requested driver")
 	driverFileName := toFilename(d, &kr, driverName, driverType)
-	destination := toLocalPath(driverVer, driverFileName, kr.Architecture.ToNonDeb())
-	if exist, _ := utils.FileExists(destination); exist {
-		return destination, ErrAlreadyPresent
+	destPath := toLocalPath(driverVer, driverFileName, kr.Architecture.ToNonDeb())
+	if exist, _ := utils.FileExists(destPath); exist {
+		return destPath, ErrAlreadyPresent
 	}
+	srcPath := fmt.Sprintf("/usr/src/%s-%s", driverName, driverVer)
```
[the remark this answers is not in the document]
FedeDP: Done! Force pushed the change.

PR #357, opened 2023-11-24
new(cmd,internal/utils,pkg/driver): use correct `engine.kind` config key

FedeDP:
**What type of PR is this?**

/kind cleanup

/area library
/area cli
/area tests

Moreover, added a new ReplaceTextInFile utils, and added tests for it and ReplaceLineInFile.

Fixes #


cmd/driver/config/config.go
```diff
@@ -169,8 +173,38 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
...
+	yamlFile, err := os.ReadFile(filepath.Clean(falcoCfgFile))
+	if err != nil {
+		return err
+	}
+	cfg := falcoCfg{}
+	if err = yaml.Unmarshal(yamlFile, &cfg); err != nil {
+		return err
+	}
+	if err = checkFalcoRunsWithDrivers(cfg.Engine.Kind); err != nil {
+		return err
```
FedeDP, line 204: I am not sure whether we should error out in this case.
ConfigMap loop does just `continue`.

pkg/driver/type/bpf.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // TypeBpf is the string for the bpf driver type.
-const TypeBpf = "bpf"
+const TypeBpf = "ebpf"
```
FedeDP, line 31: New name. See https://github.com/falcosecurity/falco/pull/2413

pkg/driver/type/modernbpf.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 // TypeModernBpf is the string for the bpf driver type.
-const TypeModernBpf = "modern-bpf"
+const TypeModernBpf = "modern_ebpf"
```
FedeDP, line 26: New name, see https://github.com/falcosecurity/falco/pull/2413

cmd/driver/config/config.go
FedeDP: I think we might not want to give an error in this case but just print a warning?
Indeed, we might want to add a `-f,--force` option to force the update.

FedeDP: I added the changes in the latest commit.
Example output when running fine and when running with the warning:
```
sudo ./falcoctl driver config --host-root "." --type ebpf
2023-11-24 12:20:02 INFO  Running falcoctl driver config driver host root: . driver type: ebpf

sudo ./falcoctl driver config --host-root "." --type ebpf
2023-11-24 12:20:09 INFO  Running falcoctl driver config driver host root: . driver type: ebpf
2023-11-24 12:20:09 WARN  Avoid updating Falco configuration config: etc/falco/falco.yaml reason: engine.kind is not driver driven: gvisor
```

FedeDP:
Rebased on main.

cmd/driver/config/config.go
```diff
@@ -161,26 +165,58 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
...
 
-func replaceDriverTypeInFalcoConfig(hostRoot string, driverType drivertype.DriverType) error {
-	return utils.ReplaceLineInFile(hostRoot+"/etc/falco/falco.yaml", "driver_mode:", "driver_mode: "+driverType.String(), 1)
+func checkFalcoRunsWithDrivers(engineKind string) error {
+	// Modify the data in the ConfigMap/Falco config file ONLY if engine.kind is set to a known driver type.
+	// This ensures that we modify the config only for Falcos running with drivers, and not plugins/gvisor.
+	// Scenario: user has multiple Falco pods deployed in its cluster, one running with driver,
+	// other running with plugins. We must only touch the one running with driver.
+	if _, err := drivertype.Parse(engineKind); err != nil {
+		return fmt.Errorf("engine.kind is not driver driven: %s", engineKind)
```
alacuku, line 182: We need a more meaningful error description. Or at least be clear in the command description that we only support `driver driven` deployments.

cmd/driver/config/config.go
```diff
@@ -161,26 +165,58 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
...
 
-func replaceDriverTypeInK8SConfigMap(ctx context.Context, namespace, kubeconfig string, driverType drivertype.DriverType) error {
+func (o *driverConfigOptions) replaceDriverTypeInFalcoConfig(hostRoot string, driverType drivertype.DriverType) error {
+	falcoCfgFile := filepath.Join(hostRoot, "etc", "falco", "falco.yaml")
+	type engineCfg struct {
+		Kind string `yaml:"kind"`
+	}
+	type falcoCfg struct {
+		Engine engineCfg `yaml:"engine"`
+	}
```
alacuku, line 194: Can we please declare these new types outside the function?

cmd/driver/config/config.go
```diff
@@ -161,26 +165,58 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
...
+	yamlFile, err := os.ReadFile(filepath.Clean(falcoCfgFile))
+	if err != nil {
+		return err
+	}
+	cfg := falcoCfg{}
+	if err = yaml.Unmarshal(yamlFile, &cfg); err != nil {
+		return err
+	}
```
alacuku, line 202: Wondering if there is a better way to do this.

alacuku (commented):
Hey @FedeDP, left some comments.

cmd/driver/config/config.go
FedeDP: We could just manually parse the file line by line, but i think this is the best way to achieve it actually.

cmd/driver/config/config.go
FedeDP: I think they are good there since they are only needed to unmarshal the Falco `engine.kind` configuration key; no need to expose them elsewhere in the package.

cmd/driver/config/config.go
FedeDP: Agree!
Was thinking about updating long description with:
```
		Long:                  "Configure a driver for future usages with other driver subcommands. Only supports deployments of Falco that use a driver engine, ie: one between kmod,ebpf and modern-ebpf.",
```

FedeDP: Better still:
```
`Configure a driver for future usages with other driver subcommands.
It will also update local Falco config or k8s configmap depending on the environment where it is running.
Only supports deployments of Falco that use a driver engine, ie: one between kmod, ebpf and modern-ebpf.
If engine.kind key is set to a non-driver driven engine, Falco configuration won't be touched. 
`
```

alacuku (approved):

cmd/driver/config/config.go
FedeDP: Pushed the updated string! Thank you Aldo!

leogr (approved):

PR #358, opened 2023-11-28
fix(cmd/driver,pkg/driver): avoid using `filepath.Join` when joining host-root

FedeDP:
**What type of PR is this?**

/kind bug

/area library

old code:
```
falcoCfgFile := filepath.Join(hostRoot, "etc", "falco", "falco.yaml")
```
* when hostRoot is empty, will resolve to `etc/falco/falco.yaml`;
* when hostRoot is ".", will resolve to `etc/falco/falco.yaml`

Instead, we want `/etc/falco/falco.yaml` and `./etc/falco/falco.yaml`.
Just use `+`.

Fixes #


cmd/driver/config/config.go
```diff
@@ -190,7 +190,7 @@ func checkFalcoRunsWithDrivers(engineKind string) error {
 }
 
 func (o *driverConfigOptions) replaceDriverTypeInFalcoConfig(hostRoot string, driverType drivertype.DriverType) error {
-	falcoCfgFile := filepath.Join(hostRoot, "etc", "falco", "falco.yaml")
+	falcoCfgFile := hostRoot + "/etc/falco/falco.yaml"
```
leogr, line 193:
```suggestion
	falcoCfgFile := filepath.Clean(filepath.Join(os.PathSeparator, hostRoot, "etc", "falco", "falco.yaml"))
```

leogr (commented):
What about using https://pkg.go.dev/path/filepath#Clean instead?

cmd/driver/config/config.go
FedeDP: I am not sure what would then happen if you pass eg: `.` as hostRoot; we would end up with `/./etc/falco/falco.yaml` that is != than `./etc/falco/falco.yaml`.

FedeDP: (this is what we do in tests btw)

FedeDP: Also, we use string concatenation with hostRoot everywhere hostRoot is used in `pkg/driver`. I would go in with this PR and eventually change all of them to an unified method (perhaps a shared util function?) in a new one.

leogr: ok got it.
I don't like too much the idea of using string concatenation in paths. But, I will be ok with this if there's no alternative. @alacuku, any idea? :thinking:

alacuku: > I am not sure what would then happen if you pass eg: `.` as hostRoot; we would end up with `/./etc/falco/falco.yaml` that is != than `./etc/falco/falco.yaml`.

Whether `hostRoot` is defined or not we should always get an absolute path. The problem here is in the tests where we are not using an absolute path. I would change the tests in this case, and make sure that `hostRoot` has a valid value before joining the path.

FedeDP: > and make sure that hostRoot has a valid value before joining the path.

It is better to introduce an utility function and use it everywhere then; ok i'll do it!

cmd/driver/config/config.go
```diff
@@ -125,6 +125,9 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
 		})
 	}
 	if f := cmd.Flags().Lookup("host-root"); f != nil && f.Changed {
+		if !filepath.IsAbs(o.HostRoot) {
```
FedeDP, line 128: Must be absolute path now.

internal/config/config.go
```diff
@@ -208,7 +208,7 @@ func init() {
 		Name:     "falco",
 		Repos:    []string{"https://download.falco.org/driver"},
 		Version:  "",
-		HostRoot: "",
+		HostRoot: string(os.PathSeparator),
```
FedeDP, line 211: New default.

pkg/driver/distro/distro.go
```diff
@@ -43,40 +43,46 @@ const (
 	kernelSrcDownloadFolder = "kernel-sources"
 )
 
-var distros = map[string]Distro{}
-
-// ErrUnsupported is the error returned when the target distro is not supported.
-var ErrUnsupported = errors.New("failed to determine distro")
+var (
+	distros  = map[string]Distro{}
+	hostRoot = string(os.PathSeparator)
```
FedeDP, line 48: Package local variable to store running hostRoot.

pkg/driver/distro/distro.go
```diff
@@ -43,40 +43,46 @@ const (
...
 }
 
-// DiscoverDistro tries to fetch the correct Distro by looking at /etc/os-release or
+// Discover tries to fetch the correct Distro by looking at /etc/os-release or
 // by cycling on all supported distros and checking them one by one.
 //
 //nolint:gocritic // the method shall not be able to modify kr
-func DiscoverDistro(kr kernelrelease.KernelRelease, hostRoot string) (Distro, error) {
-	distro, err := getOSReleaseDistro(&kr, hostRoot)
+func Discover(kr kernelrelease.KernelRelease, hostroot string) (Distro, error) {
```
FedeDP, line 72: Renamed.

alacuku (approved):
/LGTM

FedeDP:
squashed to a single commit :)

leogr (approved):

PR #362, opened 2023-11-30
new(cmd/driver,pkg/options,internal/config): added `driver config` cmd tests

FedeDP:
**What type of PR is this?**

/kind cleanup

/area cli
/area tests

Fixes #

### DIFF
--- a/cmd/driver/config/config.go
+++ b/cmd/driver/config/config.go
@@ -158,7 +158,10 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
 
 			d, err := driverdistro.Discover(info, driverCfg.HostRoot)
 			if err != nil {
-				return err
+				if !errors.Is(err, driverdistro.ErrUnsupported) {
+					return err
+				}
+				o.Printer.Logger.Info("Detected an unsupported target system; falling back at generic logic.")
 			}
 			o.Printer.Logger.Debug("Discovered distro", o.Printer.Logger.Args("target", d))
 
--- a/cmd/driver/config/config_test.go
+++ b/cmd/driver/config/config_test.go
@@ -1,0 +1,82 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 The Falco Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package driverconfig_test
+
+import (
+	"regexp"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+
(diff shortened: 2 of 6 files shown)


cmd/driver/config/config.go
```diff
@@ -158,7 +158,10 @@ func (o *driverConfigOptions) RunDriverConfig(ctx context.Context, cmd *cobra.Co
 
 			d, err := driverdistro.Discover(info, driverCfg.HostRoot)
 			if err != nil {
-				return err
+				if !errors.Is(err, driverdistro.ErrUnsupported) {
```
FedeDP, line 161: When error is ErrUnsupported, just print that we detected an unsupported OS (we failed to determine it), and go on using the fallback generic logic.

cmd/driver/printenv/printenv.go
```diff
@@ -67,7 +68,9 @@ func (o *driverPrintenvOptions) RunDriverPrintenv(_ context.Context) error {
 
 	d, err := driverdistro.Discover(kr, driver.HostRoot)
 	if err != nil {
-		return err
+		if !errors.Is(err, driverdistro.ErrUnsupported) {
```
FedeDP, line 71: When error is ErrUnsupported go on with `undetermined` distro target.

internal/config/config.go
```diff
@@ -231,7 +231,11 @@ func Load(path string) error {
 	// Set default registry auth config path
 	viper.SetDefault(RegistryCredentialConfigKey, DefaultRegistryCredentialConfPath)
 	// Set default driver
-	viper.SetDefault(DriverKey, DefaultDriver)
+	viper.SetDefault(DriverTypeKey, DefaultDriver.Type)
```
FedeDP, line 234: This fixes the retrieving of `DriverFooKey` when falcoctl.yaml is an empty file (as used in tests).
Otherwise `Driverer` would find an empty `DriverTypeKey` value and fail during the conversion to a `drivertype.Type`.

internal/config/config.go
```diff
@@ -642,7 +646,7 @@ func UpdateConfigFile(key string, value interface{}, path string) error {
 	v.Set(key, value)
 
 	if err := v.WriteConfig(); err != nil {
-		return fmt.Errorf("unable to set key %q to config file: %w", IndexesKey, err)
+		return fmt.Errorf("unable to set key %q to config file: %w", key, err)
```
FedeDP, line 649: This was just a wrong log.

pkg/options/driver.go
```diff
@@ -26,7 +27,9 @@ type DriverTypes struct {
 
 // NewDriverTypes returns a new Enum configured for the driver types.
 func NewDriverTypes() *DriverTypes {
+	types := drivertype.GetTypes()
+	sort.Strings(types)
 	return &DriverTypes{
-		Enum: NewEnum(drivertype.GetTypes(), drivertype.TypeKmod),
+		Enum: NewEnum(types, drivertype.TypeKmod),
```
FedeDP, line 33: To get a stable CLI help message, we need to sort the strings so that they always appear with same sorting.

cmd/driver/config/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+//nolint:unused // false positive
+var addAssertFailedBehavior = func(specificError string) {
+	It("check that fails and the usage is not printed", func() {
+		Expect(err).To(HaveOccurred())
+		Expect(output).Should(gbytes.Say(regexp.QuoteMeta(specificError)))
+	})
+}
+
+//nolint:unused // false positive
+var driverConfigTests = Describe("config", func() {
```
alacuku, line 40:
```suggestion
var _ = Describe("config", func() {
```

This way the linter should not complain.

alacuku (approved):

leogr (approved):

PR #365, opened 2023-12-05
cleanup(cmd,internal,pkg): move `driver config` options to be common to all `driver` commands

FedeDP:
**What type of PR is this?**

/kind cleanup

/area library
/area cli

This way all driver commands can customize their run with both CLI args or env variables, without actually storing anything in the config file.

Fixes #

### DIFF
--- a/cmd/driver/driver_linux.go
+++ b/cmd/driver/driver_linux.go
@@ -22,7 +22,10 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
+	"strings"
 
+	"github.com/blang/semver"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 
@@ -76,7 +79,7 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
 				if err != nil {
 					return err
 				}
-				if err := cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val)); err != nil {
+				if err := cmd.Flags().Set(f.Name, strings.Join(val, ",")); err != nil {
 					return fmt.Errorf("unable to overwrite \"repo\" flag: %w", err)
 				}
 			}
@@ -149,6 +152,11 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
 					return fmt.Errorf("automatic driver selection failed")
 				}
 			}
+			// If empty, try to load it automatically from /usr/src sub folders,
+			// using the most recent (ie: the one with greatest semver) driver version.
+			if driver.Version == "" {
+				driver.Version = loadDriverVersion()
+			}
 			return driver.Validate()
 		},
 	}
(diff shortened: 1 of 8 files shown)


cmd/driver/install/install.go
```diff
@@ -64,16 +65,12 @@ func NewDriverInstallCmd(ctx context.Context, opt *options.Common) *cobra.Comman
...
 		RunE: func(cmd *cobra.Command, args []string) error {
-			driver, err := config.Driverer()
-			if err != nil {
-				return err
-			}
 			// If empty, try to load it automatically from /usr/src sub folders,
 			// using the most recent (ie: the one with greatest semver) driver version.
-			if driver.Version == "" {
-				driver.Version = loadDriverVersion()
+			if o.Driver.Version == "" {
```
FedeDP, line 70: I don't know if we should move this thing in `driver` command too; perhaps it makes sense after all.

pkg/options/driver.go
```diff
@@ -34,3 +37,31 @@ func NewDriverTypes() *DriverTypes {
...
+		Name:     d.Name,
+		Repos:    d.Repos,
+		Version:  d.Version,
+		HostRoot: d.HostRoot,
+	}
+}
+
+// Validate runs all validators steps for Driver options.
+func (d *Driver) Validate() error {
+	if !filepath.IsAbs(d.HostRoot) {
```
FedeDP, line 63: For now, the only invalid value might be a relative hostroot.
Perhaps we might want to test that eg: all driver repos are indeed urls?

cmd/driver/install/install.go
FedeDP: This way, we always try hard to discover the driver version when empty.

FedeDP: Done in latest commit.

pkg/options/driver.go
FedeDP: Done in latest commit.

cmd/driver/driver_linux.go
```diff
@@ -41,13 +50,118 @@ func NewDriverCmd(ctx context.Context, opt *commonoptions.Common) *cobra.Command
...
+				// should never happen
+				return fmt.Errorf("unable to retrieve flag type")
+			} else if !f.Changed && viper.IsSet(config.DriverTypeKey) {
+				val := viper.Get(config.DriverTypeKey)
+				if err := cmd.Flags().Set(f.Name, fmt.Sprintf("%v", val)); err != nil {
+					return fmt.Errorf("unable to overwrite \"type\" flag: %w", err)
+				}
+			}
+
+			if driverTypes.String() != "auto" {
```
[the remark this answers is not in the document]
FedeDP: We now support `auto` logic for all driver commands.
This can be useful to eg: test with `driver printenv` which driver is being discovered for `auto` driver type.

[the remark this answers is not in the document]
FedeDP: Or, to store driver config with correct driver built from automatic discovered one (when using `driver config --type auto`).

cmd/driver/driver_linux.go
```diff
@@ -41,13 +53,144 @@ func NewDriverCmd(ctx context.Context, opt *commonoptions.Common) *cobra.Command
...
```
leogr, line 123: `const`?

leogr: @FedeDP just nit

leogr (approved):

PR #369, opened 2023-12-06
fix(pkg,cmd): eBPF probe must be symlinked under /root/.falco/falco-bpf.o

FedeDP:
**What type of PR is this?**

/kind bug

/area library
/area cli

Also, avoid mounting /sys/kernel/debug in bpf `Cleanup` method, since it makes no sense. It is actually part of the `Build`.

Fixes #


cmd/driver/install/install.go
```diff
@@ -163,6 +163,7 @@ func (o *driverInstallOptions) RunDriverInstall(ctx context.Context) (string, er
 		// Print much more readable output as-is
 		o.Printer.DefaultText.Print(buf.String())
 	}
+	buf.Reset()
```
FedeDP, line 166: Small bug lingered in with latest commits.

leogr (approved):

PR #376, opened 2023-12-19
fix(cmd/driver): fixed loadDriverVersion implementation for non-semver driver versions

FedeDP:
**What type of PR is this?**

/kind bug

/area cli

In `loadDriverVersion` helper, we weren't managing the case when `driverVersion` is not a semver (ie: it's a development libs version).

Fixes #

### DIFF
--- a/cmd/driver/driver_linux.go
+++ b/cmd/driver/driver_linux.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/driver/driver_linux.go
```diff
@@ -177,12 +178,23 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
...
-	paths, _ := filepath.Glob("/usr/src/falco-*+driver")
+	paths, _ := filepath.Glob("/usr/src/falco-*")
 	for _, path := range paths {
+		fileInfo, err := os.Stat(path)
+		// We expect path to point to a folder,
+		// otherwise skip it.
+		if err != nil {
+			continue
+		}
+		if !fileInfo.IsDir() {
```
FedeDP, line 189: Skip non-directory paths.

cmd/driver/driver_linux.go
```diff
@@ -177,12 +178,23 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
...
+			continue
+		}
 		drvVer := strings.TrimPrefix(filepath.Base(path), "falco-")
 		sv, err := semver.Parse(drvVer)
 		if err != nil {
-			continue
+			// Not a semver; return it because we
+			// Won't be able to check it against semver driver versions.
+			return drvVer
```
FedeDP, line 197: In case found driver version is not semver, we cannot check it against any other semver driver version.
Return it as is.

FedeDP: In any case, it is not expected for the user to have multiple `falco-*` folders under `/usr/src`.

cmd/driver/driver_linux.go
```diff
@@ -177,12 +178,23 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
 func loadDriverVersion() string {
 	isSet := false
 	greatestVrs := semver.Version{}
-	paths, _ := filepath.Glob("/usr/src/falco-*+driver")
+	paths, _ := filepath.Glob("/usr/src/falco-*")
```
FedeDP, line 181: Removed `+driver` suffix.

leogr (approved):

PR #416, opened 2024-02-03
feat: adding support for KMS in cosign

### LINKED ISSUE #380: Support KMS keys in cosign signature
Cosign supports [KMS providers](https://docs.sigstore.dev/key_management/overview/), the CLI command would look like:

```bash
cosign verify --key <some provider>://<some key> $IMAGE_DIGEST
```

We would need to add the `key` (or `public-key`) parameter to the Signature.cosign section to support KMS providers instead of `certificate-{oidc,identity}-*` parameters

**Why is this needed**:  
Our use case is to verify rules signed with AWS KMS and stored in AWS ECR but that would also enable the support for GCP, Azure and Vault KMS

brennoo:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

 /kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

This PR adds support for KMS providers in the cosign signature verify process, making it possible to use cloud providers KMS, hashicorp vault and kubernetes.

Fixes #380

I've tested `key` parameter with different values including `http://`, `awskms://` and local file `pubkey.pem` - it was needed to add the `private-infrastructure` flag for cases where the transaction is not sent to the transparency log


pkg/index/index/index.go
```diff
@@ -63,6 +63,8 @@ type CosignSignature struct {
 	CertificateIdentity         string `yaml:"certificate-identity"`
 	CertificateIdentityRegexp   string `yaml:"certificate-identity-regexp"`
 	CertificateGithubWorkflow   string `yaml:"certificate-github-workflow"`
+	KeyRef                      string `yaml:"key"`
+	PrivateInfrastructure       bool   `yaml:"private-infrastructure"`
```
cpanato, line 67: this should be the same name as the struct `IgnoreTlog`

```suggestion
	IgnoreTlog       bool   `yaml:"ignore-tlog"`
```

cpanato (approved):
thank you

/lgtm
/approve

PR #425, opened 2024-02-07
pkg/driver/flatcar: Update the script

krnowak:
**What type of PR is this?**

/kind bug
/kind cleanup

/area library

Fixes building kernel module on Flatcar. The bash script that was relocating the kernel tools had some issues.

It was taken straight from the old driver loader bash script, but now it's a separate script, so some constructs like `local` make no sense there. A result was that no kernel tools were patched, thus they could fail to run if they were built against a newer glibc than the one provided by the falco-driver-loader image.

Also print the output of the script unconditionally, so we may have some idea about what's going on here.

I hope that hardcoding `/host` as host root is fine - I can't see it being provided for customization purposes. Using HOST_ROOT env var wouldn't work if it was provided through `--host-root` flag.

### DIFF
--- a/pkg/driver/distro/flatcar.go
+++ b/pkg/driver/distro/flatcar.go
@@ -28,29 +28,57 @@ import (
 )
 
 const flatcarRelocateScript = `
-local -a tools=(
-		scripts/basic/fixdep
-		scripts/mod/modpost
-		tools/objtool/objtool
-	)
-local -r hostld=$(ls /host/usr/lib64/ld-linux-*.so.*)
-local -r kdir=/lib/modules/$(ls /lib/modules/)/build
-echo "** Found host dl interpreter: ${hostld}"
-for host_tool in ${tools[@]}; do
-	t=${host_tool}
-	tool=$(basename $t)
-	tool_dir=$(dirname $t)
-	host_tool=${kdir}/${host_tool}
-	if [ ! -f ${host_tool} ]; then
+set -euo pipefail
+
+shopt -s nullglob
+
+hostlds=( /host/usr/lib64/ld-linux-*.so.* )
+if [[ ${#hostlds[@]} -eq 0 ]]; then
+	echo "** no dynamic loaders found"
+	exit 1
+fi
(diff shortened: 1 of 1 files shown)


FedeDP:
/milestone v0.8.0
Thanks for this PR and fix! 
> I hope that hardcoding /host as host root is fine - I can't see it being provided for customization purposes. Using HOST_ROOT env var wouldn't work if it was provided through --host-root flag.

I'd say that's ok since this is what was already done by the old `falco-driver-loader` bash script!
Were you already able to test this on flatcar?

pkg/driver/distro/flatcar.go
```diff
@@ -94,9 +122,7 @@ func (f *flatcar) customizeBuild(ctx context.Context,
 		return nil, nil
 	}
 	printer.Logger.Info("Flatcar detected; relocating kernel tools.", printer.Logger.Args("version", f.versionID))
-	out, err := exec.CommandContext(ctx, "/bin/bash", "-c", flatcarRelocateScript).Output()
-	if err != nil {
-		printer.DefaultText.Print(string(out))
-	}
+	out, err := exec.CommandContext(ctx, "/bin/bash", "-c", flatcarRelocateScript).CombinedOutput()
+	printer.DefaultText.Print(string(out))
```
FedeDP, line 126: I'd leave it like it was before; no need to be verbose if no error happens (unformatted output is ugly :/ )

krnowak:
> /milestone v0.8.0 Thanks for this PR and fix!
> 
> > I hope that hardcoding /host as host root is fine - I can't see it being provided for customization purposes. Using HOST_ROOT env var wouldn't work if it was provided through --host-root flag.
> 
> I'd say that's ok since this is what was already done by the old `falco-driver-loader` bash script! Were you already able to test this on flatcar?

Yeah, tested it on image with glibc 2.38 and it worked - `lsmod` listed the `falco` module.

pkg/driver/distro/flatcar.go
krnowak: Done.

FedeDP (approved):
/approve

PR #484, opened 2024-03-20
new: driver selection logic

FedeDP:
**What type of PR is this?**

/kind feature
when a 

/area library
/area cli

This PR expands the driver automatic selection logic, in this way:
* `--type` CLI option/`type` config key/`FALCOCTL_DRIVER_TYPE` env var do now support multiple driver types to be specified
* if a single `type` is specified, we retain the old behavior
* when multiple `type`s are passed, the automatic selection logic will choose the best-fit given:
  * types are passed in decreasing priority order
  * driver must be supported on the machine
 
New default is to enable all drivers in this order: `modern_ebpf,ebpf,kmod`.
`COS` target does now specialize `PreferredDriver` method since it does not support `kmod`.

Fixes #

### DIFF
--- a/cmd/driver/driver_linux.go
+++ b/cmd/driver/driver_linux.go
@@ -44,7 +44,8 @@ import (
 // NewDriverCmd returns the driver command.
 func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
 	driver := &options.Driver{}
-	driverTypes := options.NewDriverTypes()
+	driverTypesEnum := options.NewDriverTypes()
+	var driverTypesStr []string
 
 	cmd := &cobra.Command{
 		Use:                   "driver",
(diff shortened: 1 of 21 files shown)


pkg/driver/type/modernbpf.go
```diff
@@ -51,6 +56,34 @@ func (m *modernBpf) HasArtifacts() bool {
 	return false
 }
 
+// Get the private symbol `probeProgram` that will be used to test for
+// type Tracing, attachType AttachTraceRawTp program availability.
+//
+//go:linkname probeProgram github.com/cilium/ebpf/features.probeProgram
+func probeProgram(spec *ebpf.ProgramSpec) error
```
FedeDP, line 63: To avoid fully rewriting the function, we "unsafely" import it using `go:linkname`.
I think this is better than copying the function locally, because then we get any change `probeProgram` might get with new versions, and if `probeProgram` changes signature/name, our build will fail while bumping the library dep.
This way, we always know if something has changed.

pkg/driver/type/modernbpf.go
```diff
@@ -51,6 +56,34 @@ func (m *modernBpf) HasArtifacts() bool {
...
+
+//nolint:gocritic // the method shall not be able to modify kr
+func (m *modernBpf) Supported(_ kernelrelease.KernelRelease) bool {
+	// We can't directly use this because it uses the wrong attachtype.
+	// err := features.HaveProgramType(ebpf.Tracing)
+	// Therefore, we need to manually build a feature test.
+	// Empty tracing program that just returns 0
+	progSpec := &ebpf.ProgramSpec{
```
FedeDP, line 71: Pass a prog spec with correct `AttachType` type.

cmd/driver/driver_linux.go
```diff
@@ -162,11 +183,22 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
 		},
 	}
 
-	cmd.PersistentFlags().Var(driverTypes, "type", "Driver type to be used "+driverTypes.Allowed())
+	cmd.PersistentFlags().StringSliceVar(&driverTypesStr, "type", config.DefaultDriver.Type,
```
FedeDP, line 186: `type` is now a slice.

cmd/driver/driver_linux.go
```diff
@@ -162,11 +183,22 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
...
+		"Driver types allowed in descending priority order "+driverTypesEnum.Allowed())
 	cmd.PersistentFlags().StringVar(&driver.Version, "version", config.DefaultDriver.Version, "Driver version to be used.")
 	cmd.PersistentFlags().StringSliceVar(&driver.Repos, "repo", config.DefaultDriver.Repos, "Driver repo to be used.")
 	cmd.PersistentFlags().StringVar(&driver.Name, "name", config.DefaultDriver.Name, "Driver name to be used.")
 	cmd.PersistentFlags().StringVar(&driver.HostRoot, "host-root", config.DefaultDriver.HostRoot, "Driver host root to be used.")
+	cmd.PersistentFlags().StringVar(&driverKernelRelease,
```
FedeDP, line 192: Moved `--kernelrelease` and `--kernelversion` CLI options directly to `driver` root command.

cmd/driver/driver_linux.go
```diff
@@ -115,44 +120,60 @@ func NewDriverCmd(ctx context.Context, opt *options.Common) *cobra.Command {
...
-				opt.Printer.Logger.Debug("Discovered distro", opt.Printer.Logger.Args("target", d))
+				allowedDriverTypes = append(allowedDriverTypes, drvType)
+			}
 
-				driver.Type = d.PreferredDriver(info)
-				if driver.Type == nil {
-					return fmt.Errorf("automatic driver selection failed")
+			// Step 2: fetch system info (kernel release/version and distro)
+			var err error
+			driver.Kr, err = driverkernel.FetchInfo(driverKernelRelease, driverKernelVersion)
```
FedeDP, line 146: `options.Driver` now holds run kernelrelease and distro too.
This way, we discover them just once, here, and then pass the driver options down to all driver related commands, without needing to discover distro or kernel infos twice.

cmd/driver/install/install.go
```diff
@@ -136,17 +119,13 @@ func (o *driverInstallOptions) RunDriverInstall(ctx context.Context) (string, er
...
-	d, err := driverdistro.Discover(kr, o.Driver.HostRoot)
-	if err != nil {
-		if errors.Is(err, driverdistro.ErrUnsupported) && o.Compile {
-			o.Download = false
-			o.Printer.Logger.Info(
-				"Detected an unsupported target system, please get in touch with the Falco community. Trying to compile anyway.")
-		} else {
-			return "", fmt.Errorf("detected an unsupported target system, please get in touch with the Falco community")
-		}
+	if o.Distro.String() == driverdistro.UndeterminedDistro && o.Compile {
```
FedeDP, line 122: Check is now a bit different because we are not discovering the distro here anymore; instead we receive a distro from `driver` root command.

We need to check if `ErrUnsupported` was returned, ie: if the distro is `UndeterminedDistro`.

pkg/driver/distro/generic.go
```diff
@@ -68,16 +67,11 @@ func (g *generic) customizeBuild(_ context.Context,
...
-	// Deadly simple default automatic selection
-	var dType drivertype.DriverType
-	switch {
-	case kr.GTE(semver.MustParse("5.8.0")):
-		dType, _ = drivertype.Parse(drivertype.TypeModernBpf)
-	case kr.SupportsProbe():
-		dType, _ = drivertype.Parse(drivertype.TypeBpf)
-	default:
-		dType, _ = drivertype.Parse(drivertype.TypeKmod)
+func (g *generic) PreferredDriver(kr kernelrelease.KernelRelease, allowedDriverTypes []drivertype.DriverType) drivertype.DriverType {
```
FedeDP, line 70: Also, i plan to reuse the `drivertype.Supported()` logic in https://github.com/falcosecurity/dbg-go, by adding a new option like: `--skip-modernbpf-supported` (that will initially default to false).
In this way, once we enable this new flag, we will start generating configs only for kernels where the modern bpf cannot run. 
Probably Falco 1.0.0 will be the right time to enable the flag.

cc @leogr wdyt?

FedeDP: Here is the PR: https://github.com/falcosecurity/dbg-go/pull/43

FedeDP: Uh that's wrong, since the PreferredDriver function for modern bpf directly checks running kernel, thus we cannot rely on it against any kernel release version.

pkg/driver/type/modernbpf.go
```diff
@@ -51,6 +56,31 @@ func (m *modernBpf) HasArtifacts() bool {
...
 }
 
+// Get the private symbol `probeProgram` that will be used to test for
+// type Tracing, attachType AttachTraceRawTp program availability.
+//
+//go:linkname probeProgram github.com/cilium/ebpf/features.probeProgram
+func probeProgram(spec *ebpf.ProgramSpec) error
+
+//nolint:gocritic // the method shall not be able to modify kr
+func (m *modernBpf) Supported(_ kernelrelease.KernelRelease) bool {
```
FedeDP, line 66: Note: modern bpf will check directly against running kernel, to see if all required features are supported.
This means that we cannot check against "foreign" kernel releases.

leogr (commented):
First quick look. I really like the new UX! :star_struck: 

It SGTM already, still I'd like to get a second opinion before approving. 
cc @alacuku

pkg/driver/kernel/kernel_others.go
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2023 The Falco Authors
```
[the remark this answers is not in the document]
FedeDP: I will bump the year on all touched files!

leogr (approved):

PR #26, opened 2019-10-21
Adding TLS generation

kris-nova:
validation

Signed-off-by: Kris Nova <kris@nivenly.com>

Adding first work for TLS in Falcoctl

Hello: I am Kris and I just wanted to do something
Signed-off-by: Kris Nova <kris@nivenly.com>

**What type of PR is this?**

> Uncomment one (or more) `/kind <>` lines:

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

**Any specific area of the project related to this PR?**

> Uncomment one (or more) `/area <>` lines:

> /area library

/area cli

> /area tests

> /area examples

**What this PR does / why we need it**:

Because it should be easy to run gRPC outputs with Falco!

**Which issue(s) this PR fixes**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change?**:

```release-note
Adding initial TLS cert material generation for gRPC outputs
```

PR #499, opened 2024-03-29
fix(pkg/driver): fixup kernel headers download

FedeDP:
**What type of PR is this?**

/kind bug
/kind cleanup

/area library

This PR fixes the templating of kernel headers download script from driverkit, that needs a fixed up kernelversion (only ubuntu uses it in driverkit, and it must be actually fixed, so that eg: `#26~22.04.1 Ubuntu...`  becomes `26~22.04.1`).

~~Moreover, it bumps driverkit to what will become v0.18.2 that contains a fix for new KERNELDIR mechanism (https://github.com/falcosecurity/driverkit/pull/330).
`wip` until driverkit v0.18.2 is released.~~ Already in falcoctl main!

Fixes #

### DIFF
--- a/pkg/driver/distro/distro.go
+++ b/pkg/driver/distro/distro.go
@@ -17,6 +17,8 @@
 package driverdistro
 
 import (
+	"bufio"
+	"bytes"
 	"compress/gzip"
 	"errors"
 	"fmt"
@@ -140,8 +142,18 @@ func loadKernelHeadersFromDk(distro string, kr kernelrelease.KernelRelease) (str
 	if err != nil {
 		return "", nil, err
 	}
+	script += "\necho $KERNELDIR"
 	out, err := exec.Command("bash", "-c", script).Output() //nolint:gosec // false positive
-	path := strings.TrimSuffix(string(out), "\n")
+	var path string
+	if err == nil {
+		// Scan all stdout line by line and
+		// store last line as KERNELDIR path.
+		reader := bytes.NewReader(out)
+		scanner := bufio.NewScanner(reader)
+		for scanner.Scan() {
+			path = scanner.Text()
+		}
+	}
 	return path, func() {
 		_ = os.RemoveAll("/tmp/kernel-download")
 		_ = os.RemoveAll(path)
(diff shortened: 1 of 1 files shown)


pkg/driver/distro/distro.go
```diff
@@ -140,8 +142,18 @@ func loadKernelHeadersFromDk(distro string, kr kernelrelease.KernelRelease) (str
 	if err != nil {
 		return "", nil, err
 	}
+	script += "\necho $KERNELDIR"
 	out, err := exec.Command("bash", "-c", script).Output() //nolint:gosec // false positive
-	path := strings.TrimSuffix(string(out), "\n")
+	var path string
```
FedeDP, line 147: Same change as https://github.com/falcosecurity/driverkit/pull/330/files#diff-65b73ea131a56dc0c89be2bdab7cfdd41144cf4549ecb9ddad27ffa3daf18d5dR78

pkg/driver/distro/distro.go
```diff
@@ -208,7 +220,12 @@ func Build(ctx context.Context,
 	// try to load kernel headers urls from driverkit.
 	if _, found := env[drivertype.KernelDirEnv]; !found {
 		printer.Logger.Debug("Trying to automatically fetch kernel headers.")
-		kernelHeadersPath, cleaner, err := loadKernelHeadersFromDk(d.String(), kr)
+		// KernelVersion needs to be fixed up; it is only used by driverkit Ubuntu builder
```
FedeDP, line 223: Same change as https://github.com/falcosecurity/falcoctl/pull/356/files#diff-2504975de5f63493638a2bee46395e0bcbdd0ff000764b454bebb5fd2731d3f1R215

FedeDP:
Bumped to driverkit v0.18.2.

pkg/driver/distro/distro.go
```diff
@@ -208,7 +220,12 @@ func Build(ctx context.Context,
...
+		// and we must ensure that it is correctly set.
+		fixedKr := d.FixupKernel(kr)
+		kVerFixedKr := kr
```
FedeDP, line 226: Don't overwrite `kr` here, instead use a copy.

FedeDP: While this is a bit ugly rn, it will be much better once https://github.com/falcosecurity/falcoctl/pull/356 lands.

PR #509, opened 2024-04-04
fix(install): fix missing dependency signature verification

LucaGuerra:
**What type of PR is this?**

/kind bug

/area cli

When using the `install` command to download an artifact and its dependencies, sometimes signature for dependencies was not properly checked even if a signature entry was present in the index. This fixes it.

Fixes #


alacuku (approved):
/lgtm

PR #530, opened 2024-04-23
fix(pkg/driver): avoid source of crash during distro Discovery.

FedeDP:
**What type of PR is this?**

/kind bug

/area library

This fix avoids a source of crash.

Fixes #

/milestone v0.8.0


leogr (approved):
Good catch!

PR #543, opened 2024-05-01
add new boolean arg for 'registry push' --add-floating-tags, it will auto add the tags for the major and the minor versions

Issif:
**What type of PR is this?**

> /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

/kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

This PR introduces a new boolean arg for `registry push`: `--add-floating-tags`. If set, with `--version=1.2.3`, the tags `1` and `1.2` and `1.2.3` will be added automatically, it will make easier to follow the semver.

@alacuku do you think we should also add `latest` with that arg or another?

Fixes #

### DIFF
--- a/cmd/registry/push/push.go
+++ b/cmd/registry/push/push.go
@@ -52,6 +52,10 @@ Example - Push artifact "myplugin.tar.gz" of type "plugin" for multiple platform
 Example - Push artifact "myrulesfile.tar.gz" of type "rulesfile":
 	falcoctl registry push --type rulesfile --version "0.1.2" localhost:5000/myrulesfile:latest myrulesfile.tar.gz
 
+Example - Push artifact "myrulesfile.tar.gz" of type "rulesfile" with floating tags for the major and minor versions (0 and 0.1):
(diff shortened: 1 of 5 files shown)


cmd/registry/push/push.go
```diff
@@ -192,6 +197,13 @@ func (o *pushOptions) runPush(ctx context.Context, args []string) error {
 		return err
 	}
 
+	if o.AutoFloatingTags {
+		t := strings.Split(o.Version, ".")
```
alacuku, line 201: Could we validate the version and ensure that it is in semver format? Hint: https://github.com/falcosecurity/falcoctl/blob/da9865e2096f992446e243a72c78ebd1d77ed96d/internal/follower/follower.go#L352-L355

cmd/registry/push/push_test.go
alacuku, line 1: We should also add some tests on the floating tags itself.

alacuku (approved):
/LGTM

PR #551, opened 2024-05-08
fix(pkg/driver): create envmap if nil since driverkit expects it.

FedeDP:
**What type of PR is this?**

/kind bug

/area library

Fixes #


alacuku (approved):
/lgtm

PR #554, opened 2024-05-10
feat: Add --platform option to artifact/install

kpaulisse:
**What type of PR is this?**

/kind failing-test

/kind feature

/area cli

This PR adds a `--platform` CLI option to the `artifact install` command.

For example: `falcoctl artifact install --platform linux/amd64 ...`

There are various other commands that accept a `--platform` argument so I implemented this in the same way.

Reason: We are deploying Falco into environments that may be air gapped so we want to produce a single OCI image with the appropriate rule sets and plugins. We want to use `falcoctl` to download the artifacts. However, this always downloads the artifacts for the architecture that the CI system (or developer laptop) is running, and this often produces errors, e.g. "unable to find a manifest matching the given platform: darwin/arm64". With this new option it's possible to specify `--platform=linux/amd64` and have this work even on an ARM build host or a local mac environment.

Additionally, when I was building this, I noticed that the tests for `artifact install` were failing on my local Mac machine, so I added this new `--platform` CLI flag to the tests that were failing so it would align with the `linux/amd64` that is hard-coded elsewhere in the test. I marked this as solving a failing test for that reason.

N/A

N/A

### DIFF
--- a/cmd/artifact/install/constants.go
+++ b/cmd/artifact/install/constants.go
@@ -19,6 +19,9 @@ const (
 	// FlagAllowedTypes is the name of the flag to specify allowed artifact types.
 	FlagAllowedTypes = "allowed-types"
 
+	// FlagPlatform is the name of the flag to override the platform.
+	FlagPlatform = "platform"
+
 	// FlagResolveDeps is the name of the flag to enable artifact dependencies resolution.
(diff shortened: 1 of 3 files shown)


alacuku (approved):
/lgtm

PR #630, opened 2024-09-02
update(driver/config)!: use the same logic to configure engine kind for different enviroments

alacuku:
**What type of PR is this?**

 /kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /area library

/area cli

/area tests

> /area examples

When running in kubernetes, falcoctl does not update the falco config map but creates a new
specialized config file with the selected engine kind. The command expects the name of the Falco
configmap from which to read the Falco configuration. This is needed to avoid conflicts when multiple
Falco's are deployed in the same cluster and namespace. Just to let you know, these are breaking changes.

The engine kind on each node could differ based on the kernel capabilities of the node. When the engine kind is set to `auto` falcoctl will choose the best engine kind that fits the node where it is running.

https://github.com/falcosecurity/charts/issues/695
https://github.com/falcosecurity/driverkit/issues/351

Fixes #

### DIFF
--- a/cmd/driver/config/commit_test.go
+++ b/cmd/driver/config/commit_test.go
@@ -1,0 +1,326 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2024 The Falco Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


FedeDP (approved):
/approve

PR #716, opened 2025-01-14
fix(follow): handle directories in tar.gz artifacts

TiagoJMartins:
**What type of PR is this?**

/kind bug

/area cli

A bit of background on my use-case:

I've configured Falco to load all rules from a given sub-directory within `/etc/falco`, e.g.: `/etc/falco/subdir/rules1.yaml`, and packaged multiple rules files into separate `rulesfile` artifacts in `tar.gz` format.

During packaging, I'll generate the final directory structure I want and archive it, knowing that `falcoctl artifact` will extract the contents onto `/etc/falco`.

The issue is that while `falco artifact install` works great, `falco artifact follow` seems to break when processing the file path as it expects base paths to always be files, so it'll always fail when the rules gets updated and then it tries to move a directory.

This PR addresses the issue above by modifying the file path handling in a way that preserves the artifact directory structure, and also slightly modifies the tar.gz extraction logic by only appending to `files` if it's handling an actual file and not a directory/symlink.

I've added tests for these use cases to ensure that the previous behavior remains unchanged.

The error below is printed when trying to update an artifact pushed as `.tar.gz`, that includes the file: `/rules.csq.d/kubernetes.yaml`, which gets installed at `/etc/falco/rules.csq.d/kubernetes.yaml` (because `/rulesfiles` is mounted at `/etc/falco` by the Helm chart).

```json
{
    "destDirectory": "/rulesfiles",
    "fileName": "rules.csq.d",
    "followerName": "foobar/falco-rules/platform:1-kubernetes",
    "level": "ERROR",
    "msg": "Unable to move file",
    "reason": "unable to read file /tmp/falcoctl-1880591248/rules.csq.d: read /tmp/falcoctl-1880591248/rules.csq.d: is a directory",
    "timestamp": "2025-01-14 17:03:32"
}
```

### DIFF
--- a/internal/follower/follower.go
+++ b/internal/follower/follower.go
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: Apache-2.0
-// Copyright (C) 2023 The Falco Authors
+// Copyright (C) 2025 The Falco Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


TiagoJMartins:
Marked as WIP as it seems to work but I want to build the Docker image and run a few more tests on my end.

TiagoJMartins:
Tested with this image: `tiagojmartins/falcoctl:0.10.1` - [Docker Hub](https://hub.docker.com/repository/docker/tiagojmartins/falcoctl/tags/0.10.1/sha256:373560b65aecd91e63ef5652374bb7248025bef998831ec3c4cd5f196ef82845)

alacuku:
Hi @TiagoJMartins, thansk for the PR. 

Could you please have a look at the CI? Linting and tests are failing.

TiagoJMartins:
> Hi @TiagoJMartins, thansk for the PR.
> 
> Could you please have a look at the CI? Linting and tests are failing.

I think the issues should be fixed now. I'll wait until someone approves the next CI run 👍🏻

internal/follower/follower.go
```diff
@@ -114,8 +114,18 @@ func New(ref string, printer *output.Printer, conf *Config) (*Follower, error) {
...
+	// to prevent creating temporary directories inside artifact directories.
+	baseDir := conf.TmpDir
+	if baseDir != "" {
+		baseDir = filepath.Clean(baseDir)
+		if filepath.Base(baseDir) != "." {
+			baseDir = filepath.Dir(baseDir)
+		}
+	}
+
+	tmpDir, err := os.MkdirTemp(baseDir, "falcoctl-")
```
alacuku, line 128: Why do we need this? The temporary dir is given by the user, and is the same for all the followers/artifacts.

alacuku (commented):
Hey @TiagoJMartins, thanks for the contribution! 

I left some comments.

internal/follower/follower.go
```diff
@@ -203,52 +213,84 @@ func (f *Follower) follow(ctx context.Context) {
...
+			return err
+		}
+
+		f.logger.Debug("Installing file", f.logger.Args("followerName", f.ref, "path", relPath))
+		// Check if the file exists.
+		f.logger.Debug("Checking if file already exists", f.logger.Args("followerName", f.ref, "path", relPath, "directory", dstDir))
+		_, err = os.Stat(dstPath)
+		exists := err == nil
+		if err != nil && !os.IsNotExist(err) {
+			f.logger.Error("Unable to check existence for file", f.logger.Args("followerName", f.ref, "path", relPath, "reason", err.Error()))
```
alacuku, line 258: Consider using the `utils.FileExists` function.

TiagoJMartins: Thanks! Updated 👍🏻

internal/follower/follower.go
alacuku: @TiagoJMartins, could you please provide more info about this change?

TiagoJMartins: Oops, this was left over from when I thought this was the problem. Reverted it 👍🏻 

I totally missed this PR comment, sorry about that 🙏🏻

internal/follower/follower.go
```diff
@@ -86,9 +86,7 @@ type Config struct {
 	Signature *index.Signature
 }
 
-var (
-	isInt = regexp.MustCompile(`^(0|([1-9]\d*))$`)
-)
+var isInt = regexp.MustCompile(`^(0|([1-9]\d*))$`)
```
alacuku, line 89: Could you please revert this change?

alacuku (commented):
On last thing to fix and we are ready to merge it!

internal/follower/follower.go
TiagoJMartins: Updated 👍🏻

alacuku (approved):
/lgtm

PR #733, opened 2025-02-14
wip: new(falcoctl): add automatic signature verification for drivers

LucaGuerra:
**What type of PR is this?**

/kind feature

/area cli

This adds two new command line options to `falcoctl driver`
* `--pubkey` allows you to specify a public key for signature verification. By default, it is the Falco public key used to sign packages: `https://falco.org/repo/falcosecurity-packages.asc`
* `--no-verify` skips signature verification. By default signature verification is enforced.

If the signature cannot be verified and driver build is allowed, the driver will be built locally.

This PR is still WIP because signatures are not yet present in the Falco repo. I am currently working to add them.

Fixes #


internal/signature/testdata/key
```diff
@@ -0,0 +1,82 @@
+-----BEGIN PGP PRIVATE KEY BLOCK-----
```
LucaGuerra, line 1: Note for the reviewer: this key is technically unused, however it is useful to have it if in the future we want to add some other test that are signed by the same entity as the other tests.

Obviously, like the filename says, this is a test key and its sole purpose is to run these tests.

pkg/driver/distro/distro_test.go
```diff
@@ -233,3 +236,31 @@ func TestDiscoverDistro(t *testing.T) {
...
+	type testCase struct {
+		name                 string
+		downloadedDriverDest string
+		pubkey               string
+		valid                bool
+	}
+	testCases := []testCase{
+		{"valid local pubkey", "testdata/hello.txt", "testdata/key.pub", true},
+		// {"valid remote pubkey", "testdata/hello.txt", "https://raw.githubusercontent.com/LucaGuerra/falcoctl/8875a534256686a936213b8b8dc5bbf74b5febfc/internal/signature/testdata/key.pub", true},
+		// {"invalid remote pubkey", "testdata/hello.txt", "https://raw.githubusercontent.com/LucaGuerra/falcoctl/8875a534256686a936213b8b8dc5bbf74b5febfc/internal/signature/testdata/does_not_exist.pub", false},
```
LucaGuerra, line 250: Note for the reviewer: These are disabled until the PR is merged, then, we can update the URL with the right repo and commit hash

PR #869, opened 2025-12-22
fix: deduplication of artifact and dependency references

### LINKED ISSUE #868: Inefficient deduplication logic and incorrect input handling for dependency resolution
### Description
The current artifact installation logic contains a flaw in how user inputs are processed before resolving dependencies.

### Current Behavior
The system performs a deduplication step on the input arguments provided by the user. However, this clean, deduplicated list of references is not correctly utilized as the starting point (seed) for the dependency resolution process.

### Problem:
Because the dependency resolver does not strictly use the deduplicated input set, the system fails to efficiently filter out redundant artifacts early in the process. This leads to inefficient execution and logical inconsistencies, as the resolution phase operates on a set of references that may not match the intended unique list derived from the inputs.

### Expected Behavior:
The flow should be strictly sequential:

1. Resolve and deduplicate all user inputs into a unique list of "resolved" references.
2. Pass only this unique list to the dependency resolver.
3. Ensure the resolver builds the dependency tree starting from these exact references to avoid redundancy and errors.

c2ndev:
**What type of PR is this?**

/kind bug

/area cli

This PR fixes an incorrect behavior during the artifact installation phase.
During this step, `falcoctl` attempts to deduplicate artifact and dependency references to avoid multiple pulls from the container registry.
However, this deduplication currently does not work correctly.

Fixes #868 

### DIFF
--- a/cmd/artifact/install/install.go
+++ b/cmd/artifact/install/install.go
@@ -60,6 +60,10 @@ When providing just the name of the artifact, the command will search for the ar
 the configured index files, and if found, it will use the registry and repository specified
 in the indexes.
 
+Note: If multiple versions of the same artifact are specified (e.g., foo:1.0.0 foo:2.0.0),
+the command will automatically keep only the highest version and discard the others.
+A warning will be displayed when this occurs.
+
 Example - Install "latest" tag of "k8saudit-rules" artifact by relying on index metadata:
 	falcoctl artifact install k8saudit-rules
 
@@ -342,7 +346,7 @@ func (o *artifactInstallOptions) prepareArtifactList(
 	ctx context.Context,
 	artifactConfig func(ctx context.Context, ref, os, arch string) (*oci.ArtifactConfig, error),
 	args []string,
-) (arttifacts ArtifactMapType, err error) {
+) (artifacts ArtifactMapType, err error) {
 	logger := o.Printer.Logger
 	// configMap is used to avoid getting a remote config layer more than once
 	configMap := make(map[string]*oci.ArtifactConfig)
@@ -389,11 +393,11 @@ func (o *artifactInstallOptions) prepareArtifactList(
 		if existing, ok := artifactMap[config.Name]; ok {
 			// Keep the higher version
 			if ver.Compare(*existing.ver) > 0 {
-				logger.Debug("Artifact version conflict resolved",
+				logger.Warn("Multiple versions of the same artifact detected, keeping the highest version",
 					logger.Args("artifact", config.Name, "kept", ver.String(), "discarded", existing.ver.String()))
 				artifactMap[config.Name] = &ArtifactInfo{ref: ref, ver: &ver}
 			} else {
-				logger.Debug("Artifact version conflict resolved",
+				logger.Warn("Multiple versions of the same artifact detected, keeping the highest version",
 					logger.Args("artifact", config.Name, "kept", existing.ver.String(), "discarded", ver.String()))
 			}
 		} else {
--- a/README.md
+++ b/README.md
@@ -290,6 +290,10 @@ By default, if we give the name of an **artifact** it will search for the **arti
 
  > If the repositories of the **artifacts** your are trying to install are not public then you need to authenticate to the remote registry.
 
+##### Handling Multiple Versions
+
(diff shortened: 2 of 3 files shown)


cmd/artifact/install/deps.go
leogr: Before this changeset, this function took a single `artifactConfigResolver` and assumed it resolved references by calling `IndexCache.ResolveReference(ref)`.  Now we need to explicitly call `resolver` before inserting it in the map. 
This is correctly handled for primary deps (see line 173 below), but we are missing the `resolver` call here for alternative deps. 

Indeed, if we try to install `k8saudit-eks:0.2.1` together with `k8saudit-rules` (which, among alternatives, declares `k8saudit-eks:0.4.0`) we see the problem:
```
❯ ./falcoctl artifact install k8saudit-eks:0.2.1 k8saudit-rules
2025-12-23 14:01:21 INFO  Resolving dependencies ...
2025-12-23 14:01:24 ERROR unable to create new repository with ref k8saudit-eks:0.4.0: invalid reference: missing registry or repository
```

Before, the alternative dependency was correctly resolved:
```
❯ falcoctl artifact install k8saudit-eks:0.2.1 k8saudit-rules
2025-12-23 14:01:08 INFO  Resolving dependencies ...
2025-12-23 14:01:12 INFO  Installing artifacts refs: [k8saudit-eks:0.4.0 ghcr.io/falcosecurity/plugins/ruleset/k8saudit:latest json:0.7.0]
2025-12-23 14:01:12 INFO  Preparing to pull artifact ref: ghcr.io/falcosecurity/plugins/plugin/k8saudit-eks:0.4.0
...
```

cannarelladev: the decoupling between "resolver" and pull of "configResolver" is intentional, in fact the goal is to avoid that inside the `ResolveDeps` the `inRefs` will be resolved again (they are already resolved). 
Anyway the described behavior about the alternative reference it's a bug. I'm fixing it resolving properly the ref before upserting the map.

cmd/artifact/install/install.go
```diff
@@ -233,79 +234,37 @@ func (o *artifactInstallOptions) RunArtifactInstall(ctx context.Context, args []
...
+	signatures := make(map[string]*index.Signature)
 	for _, ref := range refs {
-		resolvedRef, err := o.IndexCache.ResolveReference(ref)
-		if err != nil {
-			return err
-		}
-
-		if signatures[resolvedRef] == nil {
+		if signatures[ref] == nil {
 			if sig := o.IndexCache.SignatureForIndexRef(ref); sig != nil {
```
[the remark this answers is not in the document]
cannarelladev: Fixed: the implementation now reuses an existing data structure to store all artifact names for `SignatureForIndexRef()`.
Additionally, the behavior of `falcoctl artifact install` when duplicate references are provided is not defined anywhere in the repository. Previously, this resulted in an error; now, duplicates are handled by selecting the highest version.

cmd/artifact/install/install_prepare_test.go
```diff
@@ -0,0 +1,585 @@
...
+		foo150: {Name: "foo", Version: "1.5.0"},
+	}}
+
+	artifacts, err := o.prepareArtifactList(ctx, provider.ArtifactConfig, []string{
+		"foo:1.0.0",
+		"foo:2.0.0",
+		"foo:1.5.0",
+	})
+	assert.Len(t, artifacts, 1, "should keep only one version")
+	require.NoError(t, err)
```
leogr:
```suggestion
	require.NoError(t, err)
	assert.Len(t, artifacts, 1, "should keep only one version")
```
Nit. Shall the error stop the test immediately?

c2ndev:
> > > Finally, I want only to reiterate on the fact that this PR introduces a behavioral change. For example, providing `foo:1.0.0 foo:2.0.0` produced an error before, now silently installs `2.0.0`. If possible, I would highlight this in the output logs during processing of deps and update the documentation (if any).
> > 
> > 
> > Thank you @leogr. How we would like to manage this change? In the Documentation is nowhere mentioned the previous behavior so we only need to understand if we would consider it as a breaking or not and mention it in the changelog. Meanwhile I'm highlighting it in the output logs
> 
> Since we don't have any documentation/changelog, we may add a line to the `falcoctl artifact install --help` usage message, if you agree. Otherwise just logging it would be enough atm, I guess.

OK, I agree! I will put it on the help message and as a log

leogr (approved):

PR #882, opened 2026-01-21
 fix(artifact): add config file support for stateDir option

c2ndev:
**What type of PR is this?**

/kind bug

> /kind cleanup

> /kind design

> /kind documentation

> /kind failing-test

> /kind feature

> /kind flaky-test

> /area library

/area cli

> /area tests

> /area examples

Add support for reading `stateDir` from the configuration file in both `artifact install` and `artifact follow` commands.

The `stateDir` option could only be set via the `--state-dir` CLI flag. Although the config file accepted `artifact.install.statedir` and `artifact.follow.statedir` keys, these values were never read. 

This causes issues in Kubernetes deployments where:                                                              
  - Containers run as non-root for security                                                                        
  - The default `/var/lib/falcoctl` cannot be created (permission denied)                                          
  - The Helm chart configures a writable path but falcoctl ignores it

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,12 +125,16 @@ artifact:
     refs:
     - falco-rules:0
     - my-rules:1
+    rulesfilesdir: /tmp/rules
+    pluginsdir: /tmp/plugins
+    statedir: /tmp/state
   install:
     refs:
       - cloudtrail-rules:latest
       - cloudtrail:latest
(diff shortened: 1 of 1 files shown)


leogr (approved):
Good catch!

LGTM

PR #1035, opened 2026-07-14
utils: reject symlink and hardlink entries in ExtractTarGz

arpitjain099:
**What type of PR is this?**

/kind bug

/area library

`ExtractTarGz` (`internal/utils/extract.go`) contains extracted regular files to `destDir` (the `..` check on `header.Name` plus `safeConcat` from #228), but it does not validate link targets. A hardlink target is joined as `filepath.Join(destDir, filepath.Clean(linkname))`, and `filepath.Clean` keeps a leading `../`, so the target can resolve outside `destDir`. A symlink target (`header.Linkname`) is passed to `os.Symlink` unchanged, so an absolute or `../` target is accepted. Because artifact install directories (`PluginsDir`, `RulesfilesDir`, `AssetsDir`) are reused across installs, a planted escaping symlink can be written through by a later extraction.

Following a private report and discussion with the maintainers, this hardens the extractor by rejecting symlink and hardlink entries outright. Falco artifacts do not need link entries, so refusing them is the simplest correct fix and removes the unvalidated-link-target surface entirely.

N/A. Addresses a hardening item reported privately and coordinated with the maintainers.

- The change removes the two link-creation passes and the now-unused `link` struct plus the `links` / `symlinks` slices; `ExtractTarGz` now handles directories and regular files and returns an error on any symlink or hardlink entry.
- Added `TestExtractTarGz_RejectsLinkEntries` covering both a symlink and a hardlink entry; the existing `TestExtractTarGz` and `TestExtractTarGzStripComponents` still pass.
- Possible follow-up (not in this PR): because install dirs are shared across runs and `safeConcat` is string-only, a symlink planted by an older client could still be followed on a later regular-file write. Opening regular files with `O_NOFOLLOW` (or an `lstat` per component) would close that residual. Happy to do that separately if you want it.

```release-note
Reject symlink and hardlink entries during artifact extraction to remove an unvalidated link-target path-traversal surface.
```

### DIFF
--- a/internal/utils/extract.go
+++ b/internal/utils/extract.go
@@ -28,19 +28,12 @@ import (
 	"golang.org/x/net/context"
 )
 
-type link struct {
-	Name string
-	Path string
-}
-
 // ExtractTarGz extracts a *.tar.gz compressed archive and moves its content to destDir.
 // Returns a slice containing the full path of the extracted files.
 func ExtractTarGz(ctx context.Context, gzipStream io.Reader, destDir string, stripPathComponents int) ([]string, error) {
 	var (
-		files    []string
-		links    []link
-		symlinks []link
-		err      error
+		files []string
+		err   error
 	)
 
 	// We need an absolute path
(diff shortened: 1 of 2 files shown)


internal/utils/extract.go
```diff
@@ -106,46 +99,16 @@ func ExtractTarGz(ctx context.Context, gzipStream io.Reader, destDir string, str
...
-			}
-			if name == "" {
-				continue
-			}
-
-			name = filepath.Join(destDir, filepath.Clean(name))
-			links = append(links, link{Path: path, Name: name})
-		case tar.TypeSymlink:
-			symlinks = append(symlinks, link{Path: path, Name: header.Linkname})
+		case tar.TypeSymlink, tar.TypeLink:
```
leogr, line 102: +1 on the follow-up you mentioned 🙏

Since `safeConcat` is string-only, a symlink already sitting in one of the shared install dirs (`PluginsDir`/`RulesfilesDir`/`AssetsDir`) could still be followed when a later extraction writes a regular file through it, via the `os.OpenFile` in the `TypeReg` case above. This PR removes the way such a symlink would get planted in the first place, which is the important part.

Not for this PR though 😇 May you open an issue to track the `O_NOFOLLOW`/`lstat` hardening, so we don't lose it? 🙏

leogr (approved):
LGTM 👍 

Just left a comment re the follow-up. 

PS
This PR has been reviewed with the help of https://github.com/leogr/falco-expert

internal/utils/extract.go
arpitjain099: Done, opened #1036 to track the O_NOFOLLOW/lstat hardening.

PR #1037, opened 2026-07-15
utils: reject extraction through pre-existing symlinks in dest

### LINKED ISSUE #1036: Harden ExtractTarGz against following pre-existing symlinks in shared install dirs (O_NOFOLLOW/lstat)
From @leogr's review on #1035.

#1035 stops the archive from planting a symlink, but `safeConcat` is lexical, so a symlink already sitting in one of the install dirs (`PluginsDir`/`RulesfilesDir`/`AssetsDir`) still gets followed by the `os.OpenFile` in the `TypeReg` case, and the write ends up outside the dest.

Hardening: `O_NOFOLLOW` on the open so the final component isn't followed, and/or lstat the components before writing.

Not reachable through the archive itself after #1035, so this is defense in depth against a symlink that got there some other way. Filing so it doesn't get lost.

Happy to send the PR.

arpitjain099:
/kind cleanup

/area library

Follow-up to #1035, per @leogr's review comment there.

`safeConcat` is lexical, so a symlink already sitting in one of the install dirs still gets followed by the `os.OpenFile` in the `TypeReg` case, and the write ends up outside the dest. This lstat-checks each component under the dest root, and opens with `O_NOFOLLOW` on unix (build-tagged, Windows just gets the lstat path).

Test plants a symlink pointing outside dest and checks a regular file entry doesn't get written through it, both when the symlink is the leaf and when it's a parent dir. The link-entry rejection from #1035 is untouched.

Defense in depth really, since after #1035 you can't get such a symlink in through the archive anyway.

Fixes #1036

**Notes**:

Went with lstat plus `O_NOFOLLOW` instead of pulling in securejoin, keeps the diff small and works on all three platforms. lstat alone has a TOCTOU window, hence the `O_NOFOLLOW` on unix. Happy to switch if you'd rather.

```release-note
utils: reject extracting a regular file through a pre-existing symlink in the destination directory
```

### DIFF
--- a/internal/utils/extract.go
+++ b/internal/utils/extract.go
@@ -91,10 +91,10 @@ func ExtractTarGz(ctx context.Context, gzipStream io.Reader, destDir string, str
 			// under destDir, is a pre-existing symlink, so extraction cannot be
 			// redirected outside destDir. On unix O_NOFOLLOW below also stops the
 			// final component from being followed race-free.
-			if err = checkNoSymlinkTraversal(destDir, path); err != nil {
+			if err := checkNoSymlinkTraversal(destDir, path); err != nil {
 				return nil, err
 			}
-			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC|oNoFollow, info.Mode())
+			outFile, err := os.OpenFile(filepath.Clean(path), os.O_RDWR|os.O_CREATE|os.O_TRUNC|oNoFollow, info.Mode())
 			if err != nil {
 				return nil, err
 			}
--- a/internal/utils/extract.go
+++ b/internal/utils/extract.go
@@ -86,7 +86,15 @@ func ExtractTarGz(ctx context.Context, gzipStream io.Reader, destDir string, str
 				return nil, err
 			}
 		case tar.TypeReg:
-			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, info.Mode())
+			// safeConcat is a lexical join, so it does not resolve symlinks that
+			// already exist on disk. Reject a write whose target, or any parent
+			// under destDir, is a pre-existing symlink, so extraction cannot be
+			// redirected outside destDir. On unix O_NOFOLLOW below also stops the
+			// final component from being followed race-free.
+			if err = checkNoSymlinkTraversal(destDir, path); err != nil {
+				return nil, err
+			}
+			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC|oNoFollow, info.Mode())
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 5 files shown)


internal/utils/extract.go
```diff
@@ -86,7 +86,15 @@ func ExtractTarGz(ctx context.Context, gzipStream io.Reader, destDir string, str
 				return nil, err
 			}
 		case tar.TypeReg:
-			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC, info.Mode())
+			// safeConcat is a lexical join, so it does not resolve symlinks that
+			// already exist on disk. Reject a write whose target, or any parent
+			// under destDir, is a pre-existing symlink, so extraction cannot be
+			// redirected outside destDir. On unix O_NOFOLLOW below also stops the
+			// final component from being followed race-free.
+			if err = checkNoSymlinkTraversal(destDir, path); err != nil {
```
leogr:
```suggestion
			if err := checkNoSymlinkTraversal(destDir, path); err != nil {
```

`gocritic` (`sloppyReassign`) is failing CI here. Scoping it with `err :=` fixes it, and it matches the shadowed `err` on the `os.OpenFile` line just below.

internal/utils/extract.go
```diff
@@ -86,7 +86,15 @@ func ExtractTarGz(ctx context.Context, gzipStream io.Reader, destDir string, str
...
+				return nil, err
+			}
+			outFile, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_TRUNC|oNoFollow, info.Mode())
```
leogr:
```suggestion
			outFile, err := os.OpenFile(filepath.Clean(path), os.O_RDWR|os.O_CREATE|os.O_TRUNC|oNoFollow, info.Mode())
```

`gosec` (`G304`) fires here since the PR touches this line. Wrapping in `filepath.Clean` is how the rest of the repo silences it (e.g. `internal/utils/file.go`, `pkg/index/index/index.go`). `path` is already the validated safe join, so this is only to satisfy the linter.

A bit ironic given the whole point is that lexical cleaning isn't enough :sweat_smile: If you'd rather be explicit, `//nolint:gosec // path validated by safeConcat + checkNoSymlinkTraversal` works too. Either is fine, whatever gets CI green.

internal/utils/extract_nofollow_unix.go
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2024 The Falco Authors
```
leogr:
```suggestion
// Copyright (C) 2026 The Falco Authors
```

`goheader` wants the current year here (CI: `Expected:2026, Actual: 2024`).

internal/utils/extract_nofollow_other.go
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (C) 2024 The Falco Authors
```
leogr:
```suggestion
// Copyright (C) 2026 The Falco Authors
```

Same year fix here for consistency. CI doesn't flag this one since the Linux lint build-excludes the `!unix` file, but it should match.

N.B. tiny nit while we're here: the repo names the catch-all file `_others.go` (plural) elsewhere (e.g. `pkg/driver/kernel/kernel_others.go`). Not a blocker.

leogr (commented):
Hey @arpitjain099

Thanks for picking this up so quickly :pray: 

On your securejoin question: I would stick with what you have. Since this is defense-in-depth (after #1035 the archive cannot plant the symlink anyway), the lightweight guard is proportionate. And AFAIK `securejoin.SecureJoin` is itself lexical + `lstat`, so it would not close the residual parent-component TOCTOU either. The only thing that does is `openat2(RESOLVE_NO_SYMLINKS)` or an `openat`-based `OpenInRoot`, which is overkill here. So no need to pull it in.

The only blocker is CI is red. `golangci-lint` fails on three small things (`sloppyReassign`, `gosec`/`G304`, and `goheader` on the year). Left inline suggestions for each.

One broader thought, out of scope here: this hardens the **install** path (`ExtractTarGz` writes straight into the shared dir). The **follow** path lands via `moveFiles` then `utils.Move` (`os.Rename`, or `os.WriteFile` on the cross-device fallback), which has the same pre-existing-symlink exposure and is not covered. The `TypeDir` / `MkdirAll` case can also still create empty dirs through a pre-existing parent symlink (low severity, no content written).

Otherwise, this LGTM once CI is green :+1:

Thanks :pray:

PS
Reviewed using https://github.com/leogr/falco-expert

arpitjain099:
Thanks, all three lint fixes are pushed. Went with `filepath.Clean` over the nolint since it's what `file.go` and `index.go` already do, irony noted :sweat_smile: Took the `_others.go` rename too.

Agreed on securejoin. It's lexical plus `lstat` under the hood, so it'd buy the same guarantee I already have, and `openat2` is more machinery than a defense-in-depth guard needs here.

The follow path is a real gap though. `utils.Move` lands via `os.Rename` or the `os.WriteFile` fallback and neither goes through the guard, so a pre-existing symlink sitting in the shared dir would still get followed there. Same story for `MkdirAll` on `TypeDir`, though that one only ever creates empty dirs. I'll add both to #1036 so they don't get lost, and happy to send a follow-up for the Move path if you want it.

internal/utils/extract.go
arpitjain099: Took `filepath.Clean`, since `file.go` and `index.go` already do it that way. Left the comment above it intact so the next reader doesn't come away thinking the lexical clean is what's doing the work.

internal/utils/extract_nofollow_other.go
arpitjain099: Renamed to `extract_nofollow_others.go` to match `kernel_others.go`. Year fixed on both files.

leogr (approved):
LGTM now!

Thanks 🙏

PR #1046, opened 2026-07-31
Skip index entries without a platform in RawManifest

arpitjain099:
**What type of PR is this?**

/kind bug

/area library

`RawManifest` resolves an image index by walking its manifests:

```go
for _, manifest := range index.Manifests {
    if manifest.Platform.OS == os &&
        manifest.Platform.Architecture == arch {
```

`platform` is optional in the OCI image index schema, and `Descriptor.Platform` is a pointer, so an index carrying an entry without one, an attestation or a referrers entry for example, gives a nil dereference:

```
panic: runtime error: invalid memory address or nil pointer dereference
```

The index JSON comes from the registry via `repo.FetchReference`, and the loop runs right after unmarshalling, before any blob is fetched. `RawManifest` is also the single sink for `RawConfigLayer` and `CheckAllowedType`, so everything that resolves an artifact goes through it, including the long running `artifact follow` loop that Falco deployments use for rule and plugin updates. A crash there stops the update mechanism.

The fix skips entries without a platform, so the match falls through to the existing "unable to find a manifest matching the given platform" error when nothing matches. It matches the guards already in this file, `len(Layers) < 1` in `manifestFromDesc` and `len == 0` in `CheckAllowedType`.

None

Added a standalone test that stands up an httptest registry serving an index whose first entry has no platform and whose second is linux/amd64. It panics on main and passes with the change.

The existing Ginkgo `TestPuller` suite needs a local registry, so it fails the same way on an unmodified tree here; I did not touch it. `go test ./pkg/oci/pusher/` is green.

I am happy to take this privately instead if you would rather, given `artifact follow` runs continuously. It looked like ordinary crash hardening to me since the fix is a nil check on optional spec data, but your call.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/mitchellh/mapstructure v1.5.1-0.20231216201459-8508981c8b6c
 	github.com/onsi/ginkgo/v2 v2.32.0
 	github.com/onsi/gomega v1.42.1
+	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/pterm/pterm v0.12.82
 	github.com/robfig/cron/v3 v3.0.1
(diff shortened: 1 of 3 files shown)


alacuku (approved):
/lgtm

PR #1089, opened 2026-09-14
fix(driver): preserve DKMS installations during driver install

leogr:
**What type of PR is this?**

/kind bug

/area cli
/area tests

Running `falcoctl driver install` with a cached module removes its DKMS installation, then reports success after `insmod`. A later `modprobe`, including Falco's systemd startup, fails with `Module falco not found`. `--download=false` has the same problem, since the build path also returns early on a cached artifact.

Preserve DKMS installations and unload the existing module only when its replacement is ready. Local compilation always reaches DKMS and verifies that the requested module is installed. Missing sources and loading failures return an error.

Found during Falco 0.45.0-rc2 package testing; the same installer behavior is present in falcoctl 0.13.0 and 0.14.1.

Related package lifecycle fixes: https://github.com/falcosecurity/falco/pull/3994

- Downloaded drivers keep the existing transient loading behavior used by driver-loader containers. Native installation can use `--download=false` to ensure persistence.
- Explicit `driver cleanup` still removes DKMS installations.
- Selecting another kernel acquires its driver without attempting to load it into the running kernel.
- Regression tests cover the actual install command, cleanup and DKMS status handling. The driver suite, `go vet ./...` and the repository's diff-based lint check pass.
- The full suite has one unrelated registry test failure in this environment: `noregistry` resolves to loopback instead of failing DNS lookup. The same failure reproduces on pristine `83bd81360`.
- Ubuntu 6.8 VM checks confirm cached reinstalls preserve DKMS and systemd startup, while `--download=false` restores persistence after explicit cleanup. Tested drivers 10.2.0+driver and 11.0.0-rc1+driver, including startup after reboot.

```release-note
Preserve DKMS installations when installing cached or downloaded kernel modules. Ensure local compilation verifies persistent installation, and report module loading failures.
```


c2ndev (approved):
/lgtm

